PR #1, opened 2025-05-12
Clean up the repo

ferranbt:
## 📝 Summary

This PR cleans up the repo from rbuilder references and fixes compilation issues.


.github/workflows/release.yaml
avalonche, line 1: we can rename release-op.yaml to release.yaml?

Dockerfile
avalonche, line 1: same here, can rename op-rbuilder docker file to `Dockerfile`

PR #6, opened 2025-05-12
Fix mem leak and pending pool

SozinM:
## 📝 Summary

Rebase on latest reth to include all fixes (for pending pool and memory leak) 

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/integration/integration_test.rs
+++ b/crates/op-rbuilder/src/integration/integration_test.rs
@@ -10,11 +10,11 @@ mod tests {
     use alloy_consensus::{Transaction, TxEip1559};
     use alloy_eips::{eip1559::MIN_PROTOCOL_BASE_FEE, eip2718::Encodable2718};
     use alloy_primitives::hex;
-    use alloy_provider::{Identity, Provider, ProviderBuilder};
+    use alloy_provider::{ext::TxPoolApi, Identity, Provider, ProviderBuilder};
     use alloy_rpc_types_eth::BlockTransactionsKind;
     use futures_util::StreamExt;
     use op_alloy_consensus::OpTypedTransaction;
-    use op_alloy_network::Optimism;
+    use op_alloy_network::{Optimism, TransactionResponse};
     use std::{
         cmp::max,
         path::PathBuf,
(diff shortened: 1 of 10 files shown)


crates/op-rbuilder/src/integration/integration_test.rs
```diff
@@ -99,6 +104,7 @@ mod tests {
         Ok(())
     }
 
+    /// Caution: this test could be brittle to change because of queued pool checks
```
avalonche, line 107: what queued pool checks here would make the tests brittle?

crates/op-rbuilder/src/integration/integration_test.rs
```diff
@@ -144,68 +150,135 @@ mod tests {
...
             latest_block.header.base_fee_per_gas.unwrap(),
             MIN_PROTOCOL_BASE_FEE,
         );
         for _ in 0..10 {
+            // We check queued pool to see if it already has tx we sent earlier
+            // We do this in the beginning, because once we insert valid tx the queued tx would move
+            // to pending pool, because nonce gap would be fixed.
+            let mut queued_hash = None;
+            let mut pool_content = provider.txpool_content().await?;
+            if !pool_content.queued.is_empty() {
```
avalonche, line 165: you shouldn't need the empty check here since you already have an assertion for the length of the queued transaction later

crates/op-rbuilder/src/integration/integration_test.rs
SozinM: there asserts
```
                assert_eq!(
                    pool_content.queued.len(),
                    1,
                    "Queued pool should contain only 1 transaction"
                );
```
If you add/remove transactions in the tests this assert will fail

crates/op-rbuilder/src/integration/integration_test.rs
SozinM: this assert needed on the iteration zero, when we don't have any transaction in the pool
This will store hash that is inside pending pool in case we are on the 1+ iteration

avalonche (approved):

PR #7, opened 2025-05-12
fix: use fastlz sizes for tracking da usage

danyalprout:
## 📝 Summary

Migrated from rbuilder PR https://github.com/flashbots/rbuilder/pull/615

This PR has a couple of changes:

* Use the FastLZ size of transactions to verify per tx and per block DA limits. This matches the op-geth sequencer behaviour. See this [issue](https://github.com/flashbots/rbuilder/issues/613) for more details on how op-geth behaves. 
* Track cumulative da bytes used 

## 💡 Motivation and Context

Without this change:

* The DA limiting is based on the uncompressed transaction size, which will result in needlessly throttling very compressible transactions.
* Throttling does not take into account the amount of bytes currently being used in the block

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5346,6 +5346,7 @@ dependencies = [
  "jsonrpsee",
  "metrics",
  "op-alloy-consensus",
+ "op-alloy-flz",
  "op-alloy-network",
  "op-alloy-rpc-types-engine",
  "op-revm",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,6 +128,7 @@ op-alloy-rpc-types-engine = { version = "0.14.1", default-features = false }
 op-alloy-rpc-jsonrpsee = { version = "0.14.1", default-features = false }
 op-alloy-network = { version = "0.14.1", default-features = false }
 op-alloy-consensus = { version = "0.14.1", default-features = false }
+op-alloy-flz = { version = "0.13.0", default-features = false }
 
 async-trait = { version = "0.1.83" }
 clap = { version = "4.4.3", features = ["derive", "env"] }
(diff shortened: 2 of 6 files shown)


SozinM:
I have copied implementation of DA limit tracking from reth, do they have this issue too?

danyalprout:
@SozinM I believe Reth uses the fastlz size not the txn length ([builder](https://github.com/paradigmxyz/reth/blob/main/crates/optimism/payload/src/builder.rs#L641), [txn](https://github.com/paradigmxyz/reth/blob/main/crates/optimism/txpool/src/transaction.rs#L113),  [fastlz](https://github.com/paradigmxyz/reth/blob/main/crates/optimism/txpool/src/transaction.rs#L70)).

Whereas currently we use [transaction.length() ](https://github.com/flashbots/op-rbuilder/pull/7/files#diff-d432224a339e3079cc93a48d264dd94bf17966c99bce682f27746aa95ba1bd1bL65), which I think is the uncompressed size. 

I'm not familiar with Reth, so my understanding of their implementation maybe wrong

SozinM:
IMO it's better to keep code closer to the Reth -> we could copy their approach on fastlz
Reason for this is it would be easier for us to rebase other reth changes later

SozinM:
We could just use chages from this PR https://github.com/paradigmxyz/reth/pull/16153/files from the file builder.rs
We could use already implemented trait and calculate DA size

avalonche (approved):

SozinM:
We need to import `reth_optimism_txpool::estimated_da_size::DataAvailabilitySized` so we could just call `tx.estimated_da_size()`

SozinM:
I think you could rebase on this PR, i hope we will merge it soon https://github.com/flashbots/op-rbuilder/pull/6, i solved all conflicts for the newer version
(Or if needed you could bump reth version even higher in you PR)

danyalprout:
Sure, will rebase this on main once your PR is merged.

PR #26, opened 2025-05-13
hack: bump block-build job deadline

0x416e746f6e:
## 📝 Summary

Set block-build job deadline at 30s.

## 💡 Motivation and Context

Until [this](https://github.com/ethereum-optimism/optimism/issues/15393) is implemented, we need to postpone the deadline for block-building jobs so that `getPayload` arriving after the spam of ~800 FCUs doesn't fail with "unknown payload" error.

---

* [ ] Run `make lint`
* [ ] Run `make test`


SozinM (approved):
it grows and grows

ferranbt (changes requested):
Lets use https://github.com/flashbots/op-rbuilder/pull/28 instead.

PR #31, opened 2025-05-13
OpenTelemetry support (experimental)

therustmonk:
Adds OpenTelemetry support.

Temporary depends on #6: its changes are also included here, since both PRs are based on the latest (not released) reth and #6 has bumped dependencies.

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,29 +110,6 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "0.14.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2179ba839ac532f50279f5da2a6c5047f791f03f6f808b4dfab11327b97902f"
-dependencies = [
- "alloy-eips 0.14.0",
- "alloy-primitives",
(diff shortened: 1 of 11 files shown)


crates/op-rbuilder/src/launcher.rs
```diff
@@ -0,0 +1,149 @@
+use super::{
```
ferranbt, line 1: Can we keep this launcher on main.rs?

crates/op-rbuilder/src/args/mod.rs
```diff
@@ -70,13 +70,17 @@ impl CliExt for Cli {
 
     /// Returns the type of builder implementation that the node is started with.
     /// Currently supports `Standard` and `Flashblocks` modes.
-    fn builder_mode(&self) -> BuilderMode {
+    fn builder_mode(&self) -> Option<BuilderMode> {
```
ferranbt, line 73: Can we revert this change? We want to default for now to Standard if nothing else is set.

karim-agha (approved):

PR #49, opened 2025-05-14
Add a --playground flag on op-rbuilder to start with the flags required to run the builder on playground

karim-agha:
Issue: https://github.com/flashbots/op-rbuilder/issues/9

This change adds the ability to autoconfigure op-rbuilder to run a locally deployed builder-playground.
Follow the instructions on how to start a builder playground under this repo: https://github.com/SozinM/builder-playground/tree/msozin/feat/flashblocks

Now we can use the following startup parameters for op-rbuilder:

- `./op-rbuilder node --builder.playground` This will start using the default $HOME/.playground/devnet directory
- `./op-rbuilder node --builder.playground=<PATH>` Will use the provided path as the working directory of the playground
---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args/playground.rs
+++ b/crates/op-rbuilder/src/args/playground.rs
@@ -1,0 +1,408 @@
+//! Autmoatic builder playground configuration.
+//!
+//! This module is used mostly for testing purposes. It allows op-rbuilder to
+//! automatically configure itself to run against a running op-builder playground.
+//!
+//! To setup the playground, checkout this repository:
+//!
+//!   https://github.com/SozinM/builder-playground/
+//!
+//! and then switch to the `msozin/feat/flashblocks` branch.
+//!
+//! Then run the following command:
+//!
+//!   go run main.go cook opstack --external-builder http://host.docker.internal:4444
+//!
+//! Wait until the playground is up and running, then run the following command to build
+//! op-rbuilder with flashblocks support:
+//!
+//!   cargo build --features flashblocks --bin op-rbuilder -p op-rbuilder
+//!
+//! then run the following command to start op-rbuilder against the playground:
+//!
+//!   target/debug/op-rbuilder node --builder.playground
+//!
+//! This will automatically try to detect the playground configuration and apply
+//! it to the op-rbuilder startup settings.
+//!
+//! Optionally you can specify the `--builder.playground` flag with a different
+//! directory to use. This is useful for testing against different playground
+//! configurations.
+
+use super::OpRbuilderArgs;
+use alloy_primitives::hex;
+use clap::{parser::ValueSource, CommandFactory};
+use core::{
+    net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
+    ops::Range,
+    time::Duration,
+};
+use eyre::{eyre, Result};
+use reth_cli::chainspec::ChainSpecParser;
+use reth_network_peers::TrustedPeer;
+use reth_optimism_chainspec::OpChainSpec;
+use reth_optimism_cli::{chainspec::OpChainSpecParser, commands::Commands};
+use secp256k1::SecretKey;
+use serde_json::Value;
+use std::{
+    fs::read_to_string,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
+use url::{Host, Url};
+
+/// This trait is used to extend Reth's CLI with additional functionality that
+/// populates the default values for the command line arguments when the user
+/// specifies that they want to use the playground.
+///
+/// The `--builder.playground` flag is used to populate the CLI arguments with
+/// default values for running the builder against the playground environment.
+///
+/// The values are populated from the default directory of the playground
+/// configuration, which is `$HOME/.playground/devnet/` by default.
+///
+/// Any manually specified CLI arguments by the user will override the defaults.
+pub trait CliExt {
+    /// Populates the default values for the CLI arguments when the user specifies
+    /// the `--builder.playground` flag.
+    fn populate_defaults(self) -> Self;
+}
+
+type Cli = reth_optimism_cli::Cli<OpChainSpecParser, OpRbuilderArgs>;
+
+impl CliExt for Cli {
+    fn populate_defaults(self) -> Self {
+        let Commands::Node(ref node_command) = self.command else {
+            // playground defaults are only relevant if running the node commands.
+            return self;
+        };
+
+        let Some(ref playground_dir) = node_command.ext.playground else {
+            // not running in playground mode.
+            return self;
+        };
+
+        let options = match PlaygroundOptions::new(playground_dir) {
+            Ok(options) => options,
+            Err(e) => exit(e),
+        };
+
+        options.apply(self)
+    }
+}
+
+struct PlaygroundOptions {
+    /// Sets node.chain in NodeCommand
+    pub chain: Arc<OpChainSpec>,
+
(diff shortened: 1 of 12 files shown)


crates/op-rbuilder/src/args/playground.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+
+fn read_docker_compose(basepath: &Path) -> Result<serde_yaml::Value> {
+    // this happens only once on statup so it's fine to read the file multiple times
+    let docker_compose = read_to_string(existing_path(basepath, "docker-compose.yaml")?)?;
+    serde_yaml::from_str(&docker_compose).map_err(|e| eyre!("Invalid docker-compose file: {e}"))
+}
+
+fn extract_service_command_flag(basepath: &Path, service: &str, flag: &str) -> Result<String> {
+    let docker_compose = read_docker_compose(basepath)?;
```
ferranbt, line 295: There is a new `manifest.json` file that should be easier to parse than the docker-compose.

SozinM:
let's just create Makefile (Justfile) command to run it?

crates/op-rbuilder/src/args/playground.rs
karim-agha: Seems like the manifest file is not created by https://github.com/SozinM/builder-playground/
Could you please point me to the repo and branch I should use that generates that file?

crates/op-rbuilder/src/primitives/reth/execution.rs
```diff
@@ -5,13 +5,6 @@ use reth_node_api::NodePrimitives;
 use reth_optimism_primitives::OpReceipt;
 use std::collections::HashSet;
 
-/// Holds the state after execution
-#[derive(Debug)]
-pub struct ExecutedPayload<N: NodePrimitives> {
```
SozinM, line 10: Finally, i was always looking at this but was hesitant to remove :))))

karim-agha:
> let's just create Makefile (Justfile) command to run it?

done

crates/op-rbuilder/src/args/playground.rs
ferranbt: I think you have the wrong repo. https://github.com/flashbots/builder-playground

justfile
```diff
@@ -0,0 +1,3 @@
+run-playground:
+  cargo build --features flashblocks --bin op-rbuilder -p op-rbuilder
```
ferranbt, line 2: Why only flashblocks?

SozinM:
Sorry, wasn't clear enough
My point is that we could do the whole thing with Makefile/Justfile, without rust code
IMO Makefile/Justfile would be easier to extend/maintain/add new cases, esp considering that devops team would be able to do it too (if they would test staff locally) 
But i'm not too opinionated about it (only a little bit)

crates/op-rbuilder/src/args/playground.rs
karim-agha: Is this manifest file the defined ports are the internal ports that are used by the services internally by the docker image rather than the publicly mapped and exposed ports. I still had to extract that info from docker compose yaml.

For example, in `manifest.json` we have:

```json
"services": [
	{
		"name": "op-geth",
		"ports": [
			{
				"name": "http",
				"port": 8545,
				"Protocol": "tcp",
				"HostPort": 0
			},
			{
				"name": "ws",
				"port": 8546,
				"Protocol": "tcp",
				"HostPort": 0
			},
			{
				"name": "authrpc",
				"port": 8551,
				"Protocol": "tcp",
				"HostPort": 0
			},
			{
				"name": "rpc",
				"port": 30303,
				"Protocol": "tcp",
				"HostPort": 0
			},
			{
				"name": "metrics",
				"port": 6061,
				"Protocol": "tcp",
				"HostPort": 0
			}
		]
	}
]
```

but then in `docker-compose.yaml` the are externally mapped to:

```yaml
  op-geth:
    ports:
    - 127.0.0.1:8546:8545
    - 127.0.0.1:8547:8546
    - 127.0.0.1:8553:8551
    - 127.0.0.1:30304:30303
    - 127.0.0.1:6061:6061
```

Do you think we should keep extracting this data the way we are doing it now or make `manifest.json` account for this port remapping?

karim-agha:
> Sorry, wasn't clear enough My point is that we could do the whole thing with Makefile/Justfile, without rust code IMO Makefile/Justfile would be easier to extend/maintain/add new cases, esp considering that devops team would be able to do it too (if they would test staff locally) But i'm not too opinionated about it (only a little bit)

No problem. I think that the logic of extracting the playground settings is too complex to be placed in a shell script. No strong opinions either, but if you don't mind I'd keep it now and revisit in a future PR when reintegrating flashblocks.

crates/op-rbuilder/src/args/playground.rs
ferranbt: The problem is that those mappings are only known as runtime after the playground is running. So I guess we have to use the docker-compose ones.

crates/op-rbuilder/src/args/playground.rs
```diff
@@ -0,0 +1,387 @@
...
+            ));
+        }
+
+        let chain = OpChainSpecParser::parse(&existing_path(path, "l2-genesis.json")?)?;
+
+        let authrpc_addr = Ipv4Addr::UNSPECIFIED.into();
+        let http_port = pick_preferred_port(2222, 3000..9999);
+        let authrpc_jwtsecret = existing_path(path, "jwtsecret")?.into();
+        let port = pick_preferred_port(30333, 30000..65535);
+        let chain_block_time = extract_chain_block_time(path)?;
```
ferranbt, line 134: Why do you want to have chain block time?

ferranbt: Is this for `flashblock_block_time`? That value is not set on playground, it is up to the builder to decide what to use.

ferranbt (approved):

ferranbt:
Can you update README on how to use it?

justfile
karim-agha: removed flashblocks from this iteration.

crates/op-rbuilder/src/args/playground.rs
karim-agha: I took it from [here](https://www.notion.so/flashbots/Commands-to-run-flashblocks-with-op-rbuilder-1f16b4a0d876806fb443c8fe110fe460), where the block time is set as part of the op-rbuilder setup?: 

```shell
./target/debug/op-rbuilder node     \
  --chain $HOME/.playground/devnet/l2-genesis.json     \
  --http \
  --http.port 2222     \
  --authrpc.port 4444 \
  --authrpc.jwtsecret $HOME/.playground/devnet/jwtsecret     \
  --port 30333 \
  --disable-discovery     \
  --metrics 127.0.0.1:9011     \
  --rollup.builder-secret-key 59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d     \
  --trusted-peers enode://3479db4d9217fb5d7a8ed4d61ac36e120b05d36c2eefb795dc42ff2e971f251a2315f5649ea1833271e020b9adc98d5db9973c7ed92d6b2f1f2223088c3d852f@127.0.0.1:30304 \
  --authrpc.addr 0.0.0.0 \
  --rollup.flashblocks-ws-url 0.0.0.0:1111 \
  --rollup.chain-block-time 2000\
```

karim-agha:
> Can you update README on how to use it?

done

PR #50, opened 2025-05-14
Have the option to use the reth payload generator

ferranbt:
## 📝 Summary

This PR enables the Reth payload generator for the vanilla block builder and introduces a new integration test that validates that under short block times the blocks still include transactions.

### DIFF
--- a/crates/op-rbuilder/src/integration/integration_test.rs
+++ b/crates/op-rbuilder/src/integration/integration_test.rs
@@ -462,6 +462,26 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(not(feature = "flashblocks"))]
+    async fn integration_test_transaction_flood_no_sleep() -> eyre::Result<()> {
+        // This test validates that if we flood the builder with many transactions
+        // and we request short block times, the builder can still eventually resolve all the transactions
+        let mut test_harness =
+            TestHarnessBuilder::new("integration_test_transaction_flood_no_sleep")
+                .build()
+                .await?;
+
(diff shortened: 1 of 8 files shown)


crates/op-rbuilder/src/payload_builder_vanilla.rs
```diff
@@ -261,6 +320,7 @@ impl<Pool, Client> OpPayloadBuilderVanilla<Pool, Client> {
         pool: Pool,
         client: Client,
         config: OpBuilderConfig,
+        enable_revert_protection: bool,
```
SozinM, line 323: i think you forgot to use this field inside execute_best_transctions

ferranbt: No, this is for the empty block, in that case there are no txns

SozinM (approved):

avalonche (approved):

crates/op-rbuilder/src/integration/integration_test.rs
```diff
@@ -461,6 +461,40 @@ mod tests {
...
+            transactions.push(tx_hash);
+        }
+
+        // After a 10 blocks all the transactions should be included in a block
+        for _ in 0..10 {
+            block_generator.submit_payload(None, 0, true).await.unwrap();
+        }
+
+        for tx in transactions {
+            provider.get_transaction_receipt(tx).await?;
```
avalonche, line 492: you don't need to check the response of the rpc here?

PR #61, opened 2025-05-21
Use correct DA transaction compression

SozinM:
## 📝 Summary

Supersedes https://github.com/flashbots/op-rbuilder/pull/7 and https://github.com/flashbots/op-rbuilder/pull/8
Added some extra integration tests for txpool

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/tests/framework/op.rs
+++ b/crates/op-rbuilder/src/tests/framework/op.rs
@@ -27,6 +27,8 @@ pub struct OpRbuilderConfig {
     chain_block_time: Option<u64>,
     flashbots_block_time: Option<u64>,
     with_revert_protection: Option<bool>,
+    namespaces: Option<String>,
+    extra_params: Option<String>,
 }
 
 impl OpRbuilderConfig {
@@ -83,6 +85,16 @@ impl OpRbuilderConfig {
         self.flashbots_block_time = Some(time);
         self
     }
+
+    pub fn with_namespaces(mut self, namespaces: Option<String>) -> Self {
+        self.namespaces = namespaces;
+        self
+    }
+
+    pub fn with_extra_params(mut self, extra_params: Option<String>) -> Self {
+        self.extra_params = extra_params;
+        self
+    }
 }
 
 impl Service for OpRbuilderConfig {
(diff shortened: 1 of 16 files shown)


SozinM:
@danyalprout please review

karim-agha (approved):

danyalprout:
I'm not sure this will work as I don't think you're incrementing `cumulative_da_bytes_used` when a transaction is included in a block, so this [check](https://github.com/flashbots/op-rbuilder/pull/61/files#diff-d432224a339e3079cc93a48d264dd94bf17966c99bce682f27746aa95ba1bd1bR63) wont be correct.

IMO it maybe worth adding a function to track this/gas used (for [example](https://github.com/flashbots/op-rbuilder/pull/7/files#diff-d432224a339e3079cc93a48d264dd94bf17966c99bce682f27746aa95ba1bd1bR91-R99)).

SozinM:
Oh, I maybe missed it :) 
Will recheck

SozinM:
@danyalprout fixed

crates/op-rbuilder/src/tests/framework/op.rs
```diff
@@ -155,6 +167,16 @@ impl Service for OpRbuilderConfig {
                 .arg(flashbots_block_time.to_string());
         }
 
+        if let Some(max_da_tx_size) = self.max_da_tx_size {
+            cmd.arg("--builder.max-da-tx-size")
```
ferranbt, line 171: Is this a flag of the builder? I though it is the batcher the one that updates these values.

danyalprout: It'll need to be on the RPC, as it changes dynamically when the batcher is congested.

If you ensure the DA Config is shared between the payload builder and the OPAddOnsBuilder it should work ([ref](https://github.com/flashbots/op-rbuilder/pull/8/files#diff-a444c4b32610ee0cee1af7d1ae5608e877cca94ecb7e4b518c38bd80951f2e40R59)) out of the box.

We're running the two closed PR's on Base Sepolia (and briefly Base Mainnet) last week and it successfully was able to throttle on both networks.

SozinM:
@ferranbt @danyalprout 
Fixed usage of DA config and updated test to use miner_setMaxDASize

danyalprout (approved):

PR #65, opened 2025-05-23
Custom engine API to share FCU between other builders

SozinM:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -51,6 +51,7 @@ pub struct OpRbuilderArgs {
     #[arg(long = "builder.enable-revert-protection", default_value = "false")]
     pub enable_revert_protection: bool,
 
+    /// Path to builder playgorund to automatically start up the node connected to it 
     #[arg(
         long = "builder.playground",
         num_args = 0..=1,
@@ -59,6 +60,8 @@ pub struct OpRbuilderArgs {
         env = "PLAYGROUND_DIR",
     )]
     pub playground: Option<PathBuf>,
+    /// List or builders in the network that FCU would be propagated to
+    pub engine_peers: Vec<String>
 }
 
 fn expand_path(s: &str) -> Result<PathBuf, String> {
--- a/crates/op-rbuilder/src/primitives/reth/engine_api_builder.rs
+++ b/crates/op-rbuilder/src/primitives/reth/engine_api_builder.rs
@@ -1,0 +1,228 @@
+//! RPC component builder
+
+pub use reth_optimism_rpc::OpEngineApi;
+use reth_node_api::{AddOnsContext, FullNodeComponents, NodeTypes};
+use reth_node_builder::rpc::{EngineApiBuilder, EngineValidatorBuilder};
+use reth_node_core::version::{CARGO_PKG_VERSION, CLIENT_CODE, VERGEN_GIT_SHA};
+use reth_optimism_node::OP_NAME_CLIENT;
+use reth_optimism_rpc::engine::OP_ENGINE_CAPABILITIES;
+use reth_optimism_rpc::OpEngineApiServer;
+use reth_payload_builder::PayloadStore;
+use reth_rpc_engine_api::EngineCapabilities;
+
+use alloy_eips::eip7685::Requests;
+use alloy_primitives::{BlockHash, B256, U64};
+use alloy_rpc_types_engine::{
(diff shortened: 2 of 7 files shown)


crates/op-rbuilder/src/args/op.rs
```diff
@@ -71,12 +74,54 @@ pub struct OpRbuilderArgs {
...
+
+    // Find the @ delimiter - it's required
+    let Some(at_pos) = s.find('@') else {
+        return Err(anyhow!(
+            "Engine peer must include JWT path after '@' (format: url@jwt_path)"
+        ));
+    };
+
+    let url_part = &s[..at_pos];
+    let jwt_path_part = &s[at_pos + 1..];
```
karim-agha, line 110: nit: this will crash the system with an unhelpful error message for this string: `url@`.  Le't make sure that `@` is not the last character.

crates/op-rbuilder/src/primitives/reth/engine_api_builder.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+impl<N, EV> EngineApiBuilder<N> for OpEngineApiBuilder<EV>
+where
+    N: FullNodeComponents<
+        Types: NodeTypes<
+            ChainSpec: EthereumHardforks,
+            Payload: EngineTypes<ExecutionData = OpExecutionData>,
+        >,
+    >,
```
karim-agha, line 90: We have now `NodeBounds` in `traits.rs`. This block of code can be simplified into: `N: NodeBounds`.

crates/op-rbuilder/src/primitives/reth/engine_api_builder.rs
```diff
@@ -0,0 +1,350 @@
...
+            let method = method.to_string();
+            let fork_choice_state = *fork_choice_state;
+            let payload_attributes = payload_attributes.clone();
+            task_executor.spawn(Box::pin(async move {
+                // For now, just log the multiplexing until we implement proper Alloy provider calls
+                let res: Result<ForkchoiceUpdated, jsonrpsee_core::client::error::Error> = client
+                    .request(method.as_str(), (fork_choice_state, payload_attributes))
+                    .await;
+                if let Err(err) = res {
+                    warn!("Failed to multiplex FCU, error: {err}");
```
karim-agha, line 186: I would also include the peer info that we've failed to multiplex to in the log warn message.

crates/op-rbuilder/src/primitives/reth/engine_api_builder.rs
```diff
@@ -0,0 +1,350 @@
...
+
+#[async_trait::async_trait]
+impl<Provider, EngineT, Pool, Validator, ChainSpec> OpEngineApiServer<EngineT>
+    for OpEngineApiExt<Provider, EngineT, Pool, Validator, ChainSpec>
+where
+    Provider: HeaderProvider + BlockReader + StateProviderFactory + 'static,
+    EngineT: EngineTypes<ExecutionData = OpExecutionData>,
+    Pool: TransactionPool + 'static,
+    Validator: EngineValidator<EngineT>,
+    ChainSpec: EthereumHardforks + Send + Sync + 'static,
```
karim-agha, line 201: Does `ChainSpec` need to be a generic parameter? As far as I'm aware we're always using `OpChainSpec`. Maybe we can reduce type complexity here by using concrete types that we know always are going to be used in op-rbuilder?

`EngineT: EngineTypes` can also be replaced with a concrete `OpEngineTypes` type.

crates/op-rbuilder/src/primitives/reth/engine_api_builder.rs
```diff
@@ -0,0 +1,350 @@
...
+            .fork_choice_updated_v1(fork_choice_state, payload_attributes)
+            .await
+    }
+
+    async fn fork_choice_updated_v2(
+        &self,
+        fork_choice_state: ForkchoiceState,
+        payload_attributes: Option<EngineT::PayloadAttributes>,
+    ) -> RpcResult<ForkchoiceUpdated> {
+        println!("fork_choice_updated_v2");
```
karim-agha, line 249:
```suggestion
        debug!!("fork_choice_updated_v2");
```

SozinM: oh. my bad, debug staff

PR #78, opened 2025-05-28
Feat/monitor event stream

ferranbt:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/src/monitor_tx_pool.rs
```diff
@@ -1,68 +1,198 @@
...
+}
+
+#[async_trait]
+impl TxpoolExtApiServer for TransactionPoolMonitorRpc {
+    fn subscribe_events(
+        &self,
+        pending_subscription_sink: PendingSubscriptionSink,
+    ) -> SubscriptionResult {
+        println!("Subscribing to txpool events");
+        let mut event_receiver = self.event_sender.subscribe();
```
ferranbt, line 95: If the buffer has elements, does this subscribe tries to read those elements first? or if will start at the beginning of the stream?

PR #93, opened 2025-06-02
Added feature-gated interop support

SozinM:
This PR adds feature-gated interop support. Right now to produce interop binary we'll need to make PR with this support.
Long term we will remove this feature once we are comfortable enabling interop for everyone.
Uses derive-more for display. 

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


avalonche (commented):
instead of a feature why not use a command line flag?

PR #96, opened 2025-06-02
Use original function with manual scaling

SozinM:
## 📝 Summary

Turned out `data_gas_fjord` was not a fitting function because it has extra logic of multiplying by 16.
I replaces it with original function, while manually dividing it by 1e6

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


karim-agha (approved):
nit: A link to something that explains why it is divided by million or more context would be helpful.

PR #97, opened 2025-06-02
Add error log in case builder tx da size sets max_da_block_size to 0

SozinM:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/standard/payload.rs
+++ b/crates/op-rbuilder/src/builders/standard/payload.rs
@@ -1,3 +1,9 @@
+use crate::{
+    builders::{generator::BuildArguments, BuilderConfig},
+    metrics::OpRBuilderMetrics,
+    primitives::reth::ExecutionInfo,
+    traits::{ClientBounds, NodeBounds, PayloadTxsBounds, PoolBounds},
+};
 use alloy_consensus::{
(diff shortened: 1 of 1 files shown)


ferranbt (approved):

crates/op-rbuilder/src/builders/standard/payload.rs
```diff
@@ -362,7 +361,13 @@ impl<Txs: PayloadTxsBounds> OpBuilder<'_, Txs> {
         let block_da_limit = ctx
             .da_config
             .max_da_block_size()
-            .map(|da_size| da_size.saturating_sub(builder_tx_da_size));
+            .map(|da_size| {
+                let da_size = da_size.saturating_sub(builder_tx_da_size);
+                if da_size == 0 {
+                    error!("Builder tx da size subtraction caused max_da_block_size to be 0. No transaction would be included.");
```
avalonche, line 367: it'll be helpful for debugging to log the builder tx da size + block da size

SozinM: I think it would be cool to add the amount of da size left at the end of the block as a metric
We already have a log if da exceeded

Issue #101, opened 2025-06-02
Not building enough Flashblocks per block


danyalprout:
Hey 👋

We’ve enabled Flashblocks on a devnet, Base Sepolia and (briefly) Base Mainnet. We’ve noticed the longer a new payload request takes on average, the less Flashblocks are built per block.

![Image](https://github.com/user-attachments/assets/339b589c-054e-4889-8bdd-29b84568b1bc)

My theory is, as rollup-boost makes a new payload request to the sequencer EL client after building Flashblocks, I believe this needs to be considered in the Flashblock timings. For example, if a network has the following properties:

- 2s blocks
- ~400ms new payload requests
- FLASHBLOCK_BLOCK_TIME=200ms (aiming for 10 flashblocks per block)

The builder will only have 1.6s and build ~8/10 Flashblocks. This impacts the amount of gas/da available per block. Right now we have a property for FLASHBLOCK_BLOCK_TIME, which we use to:

- [Derive number of flashblocks per block](https://github.com/flashbots/op-rbuilder/blob/main/crates/op-rbuilder/src/builders/flashblocks/config.rs#L52)
- [Run timer for flashblock building](https://github.com/flashbots/op-rbuilder/blob/main/crates/op-rbuilder/src/builders/flashblocks/payload.rs#L237)

I think we need to be able to independently configure these as either number of flashblocks/flashblock build time or number of flashblocks/block overhead time. Happy to send a PR if this seems reasonable

cc @ferranbt we spoke about this briefly on Friday

danyalprout:
Created a WIP [branch](https://github.com/flashbots/op-rbuilder/compare/main...danyalprout:op-rbuilder:build-duration?expand=1), going to test this today/tomorrow and verify it helps.

PR #111, opened 2025-06-04
fix: improve timing for Flashblock production

danyalprout:
## 📝 Summary

* Configure the number of Flashblocks per block and a block overhead vs. current Flashblock time. 
* Don't trigger the build_tx event, when you've built the full number of Flashblocks. This causes the function to return earlier.

## 💡 Motivation and Context

To ensure that we build the correct number of Flashblocks per block. See https://github.com/flashbots/op-rbuilder/issues/101 for full context.

I ran this on Base Sepolia today and it improved both:

1. Number of Flashblocks that were produced per block
Screenshot 2025-06-03 at 3 13 55 PM
Screenshot 2025-06-03 at 3 14 01 PM

2. Number of reorgs (where the block doesn't match the Flashblocks) dropped from ~20% to ~1%
Screenshot 2025-06-03 at 3 13 43 PM

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -93,11 +93,19 @@ pub struct FlashblocksArgs {
     )]
     pub flashblocks_addr: String,
 
-    /// flashblock block time in milliseconds
+    /// Number of Flashblocks per block
     #[arg(
-        long = "flashblocks.block-time",
-        default_value = "250",
-        env = "FLASHBLOCK_BLOCK_TIME"
+        long = "flashblocks.per-block",
+        default_value = "10",
+        env = "FLASHBLOCKS_PER_BLOCK"
     )]
-    pub flashblocks_block_time: u64,
+    pub flashblocks_per_block: u64,
+
+    /// Overhead for reset of block production (e.g. new payload request to sequencer EL)
+    #[arg(
+        long = "flashblocks.block-overhead",
+        default_value = "100",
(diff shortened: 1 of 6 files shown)


crates/op-rbuilder/src/args/op.rs
```diff
@@ -93,11 +93,19 @@ pub struct FlashblocksArgs {
...
+        long = "flashblocks.per-block",
+        default_value = "10",
+        env = "FLASHBLOCKS_PER_BLOCK"
     )]
-    pub flashblocks_block_time: u64,
+    pub flashblocks_per_block: u64,
+
+    /// Overhead for reset of block production (e.g. new payload request to sequencer EL)
+    #[arg(
+        long = "flashblocks.block-overhead",
```
avalonche, line 106: this is a constant offset but I'd imagine the overhead latency varying considerable block to block. Is there a way to calculate some offset based on block time and how close the block building starts from the end of the slot?

SozinM:
Sound good:
The general idea is following:
we use SystemTime::now in the beginning of the block building, get the time_left to build the block agains payload timestamp
Then if time_drift < 150 ms we just shrink first flashblock
 If it's more we reduce the number of flashblocks.
 Let's say it 300 ms - that mean that we should produce 1 less flashblock and shrink first flashblocks by 100ms
 If time_drift doesn't make sense (FCU arrived too late, let's say 1750 ms into the block or our system clock behaves) we probably should ignore it and hope for the best

PR #112, opened 2025-06-04
Upgrade to reth 1.4.7 and main rollup-boost branch

avalonche:
## 📝 Summary

upgrade reth to 1.4.7.

## 💡 Motivation and Context

bug fixes and compatibility with the latest rollup-boost

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/flashblocks/payload.rs
+++ b/crates/op-rbuilder/src/builders/flashblocks/payload.rs
@@ -38,7 +38,7 @@ use reth_revm::{
     State,
 };
 use revm::Database;
-use rollup_boost::primitives::{
+use rollup_boost::{
     ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashblocksPayloadV1,
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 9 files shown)


karim-agha (approved):

crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -664,7 +664,7 @@ where
             block_hash,
             transactions: new_transactions_encoded,
             withdrawals: ctx.withdrawals().cloned().unwrap_or_default().to_vec(),
-            withdrawals_root,
```
SozinM, line 667: I don't think there is correct default for this
Probably better to just expect, because it would mean that we don't support pre-isthmus chains

crates/op-rbuilder/src/tests/framework/apis.rs
```diff
@@ -2,6 +2,7 @@ use super::DEFAULT_JWT_TOKEN;
 use alloy_eips::{eip7685::Requests, BlockNumberOrTag};
 use alloy_primitives::B256;
 use alloy_rpc_types_engine::{ForkchoiceUpdated, PayloadStatus};
+use http::Uri;
```
SozinM, line 5: Have they break compatibility with uri::Uri? 
Can you do .to_string for uri::Uri?

PR #115, opened 2025-06-04
Add reverting hashes

ferranbt:
## 📝 Summary

Add reverting hashes to bundles.

### DIFF
--- a/crates/op-rbuilder/src/builders/context.rs
+++ b/crates/op-rbuilder/src/builders/context.rs
@@ -359,10 +359,16 @@ impl OpPayloadBuilderCtx {
 
         while let Some(tx) = best_txs.next(()) {
             let interop = tx.interop_deadline();
-            let exclude_reverting_txs = tx.exclude_reverting_txs();
+            let reverted_hashes = tx.reverted_hashes().clone();
+
             let tx_da_size = tx.estimated_da_size();
             let tx = tx.into_consensus();
             let tx_hash = tx.tx_hash();
+
+            // exclude reverting transaction if:
+            // - the transaction comes from a bundle and the hash **is not** in reverted hashes
+            let exclude_reverting_txs =
+                reverted_hashes.is_some() && !reverted_hashes.unwrap().contains(&tx_hash);
 
             let log_txn = |result: TxnExecutionResult| {
                 info!(target: "payload_builder", tx_hash = ?tx_hash, tx_da_size = ?tx_da_size, exclude_reverting_txs = ?exclude_reverting_txs, result = %result, "Considering transaction");
--- a/crates/op-rbuilder/src/primitives/bundle.rs
+++ b/crates/op-rbuilder/src/primitives/bundle.rs
@@ -9,6 +9,9 @@ pub struct Bundle {
     #[serde(rename = "txs")]
     pub transactions: Vec<Bytes>,
 
+    #[serde(rename = "reverting_tx_hashes")]
+    pub reverted_hashes: Vec<B256>,
+
     #[serde(rename = "maxBlockNumber")]
     pub block_number_max: Option<u64>,
 }
--- a/crates/op-rbuilder/src/revert_protection.rs
+++ b/crates/op-rbuilder/src/revert_protection.rs
@@ -128,7 +128,7 @@ where
         let mut pool_transaction: FBPooledTransaction =
             OpPooledTransaction::from_pooled(recovered).into();
 
-        pool_transaction.set_exclude_reverting_txs(true);
+        pool_transaction.set_reverted_hashes(bundle.reverted_hashes.clone());
         pool_transaction.set_conditional(bundle.conditional());
 
         let hash = self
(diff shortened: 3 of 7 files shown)


crates/op-rbuilder/src/tests/vanilla/revert.rs
```diff
@@ -123,36 +123,50 @@ async fn revert_protection_bundle() -> eyre::Result<()> {
...
     let bundle_opts = BundleOpts {
-        block_number_max: Some(4),
+        block_number_max: Some(5),
     };
 
-    let reverted_bundle = harness
+    let reverted_bundle_0 = harness
         .create_transaction()
         .with_revert()
+        .with_reverted_hash()
```
ferranbt, line 132: This is a new helper function that adds into the reverted_txn_hashes field of the bundle the hash of the transaction. Thus, it includes that txn reverted on the block.

crates/op-rbuilder/src/builders/context.rs
```diff
@@ -344,15 +345,31 @@ impl OpPayloadBuilderCtx {
...
             let log_txn = |result: TxnExecutionResult| {
                 info!(target: "payload_builder", tx_hash = ?tx_hash, tx_da_size = ?tx_da_size, exclude_reverting_txs = ?exclude_reverting_txs, result = %result, "Considering transaction");
             };
 
+            if let Some(conditional) = conditional {
+                // TODO: ideally we should get this from the txpool stream
+                if let Some(min_block_number) = conditional.block_number_min {
+                    if min_block_number > self.block_number() {
+                        continue;
+                    }
```
karim-agha: If we continue here, does the tx go back to the tx pool and get presented to us again in the future or it gets dropped and we need to keep track of it on the builder side?

crates/op-rbuilder/src/primitives/bundle.rs
```diff
@@ -9,19 +9,30 @@ pub struct Bundle {
     #[serde(rename = "txs")]
     pub transactions: Vec<Bytes>,
 
+    #[serde(rename = "reverting_tx_hashes")]
+    pub reverted_hashes: Vec<B256>,
```
karim-agha, line 13:
```suggestion
    pub reverting_hashes: Vec<B256>,
```

to make it consistent with the serde name

crates/op-rbuilder/src/revert_protection.rs
```diff
@@ -128,7 +137,9 @@ where
         let mut pool_transaction: FBPooledTransaction =
             OpPooledTransaction::from_pooled(recovered).into();
 
-        pool_transaction.set_exclude_reverting_txs(true);
+        println!("conditional: {:?}", bundle.conditional());
```
karim-agha: `debug!` or `info!`

crates/op-rbuilder/src/tests/framework/txs.rs
```diff
@@ -145,16 +153,24 @@ impl TransactionBuilder {
...
 
         if let Some(bundle_opts) = bundle_opts {
             // Send the transaction as a bundle with the bundle options
             let bundle = Bundle {
                 transactions: vec![transaction_encoded.into()],
+                reverted_hashes: if with_reverted_hash {
+                    vec![txn_hash]
+                } else {
+                    vec![]
+                },
```
karim-agha, line 171: Should the `BundleOpts` also have a field with reverting txs to be more resembling the real public API and alternatively extract that option from the bundle opts?

crates/op-rbuilder/src/tests/vanilla/revert.rs
```diff
@@ -123,42 +124,106 @@ async fn revert_protection_bundle() -> eyre::Result<()> {
...
+    assert!(block.includes(*bundle_with_min_block.tx_hash()));
+
+    // Send a bundle with a match of min and max block number
+    let bundle_with_min_and_max_block = harness
+        .create_transaction()
+        .with_revert()
+        .with_reverted_hash()
+        .with_bundle(BundleOpts {
+            block_number_max: Some(4),
+            block_number_min: Some(4),
```
karim-agha, line 213: very opinionated nit: maybe `Range<BlockNumber>` would make it more expressive because users will be able to specify it in various ways such as `a..b`, `a..`, `a..=b`, `..b`, `..=b`, etc.

crates/op-rbuilder/src/tx.rs
```diff
@@ -18,7 +18,11 @@ pub trait FBPoolTransaction: MaybeRevertingTransaction + OpPooledTx {}
 #[derive(Clone, Debug)]
 pub struct FBPooledTransaction {
     pub inner: OpPooledTransaction,
-    pub exclude_reverting_txs: bool,
+
+    /// reverted hashes for the transaction. If the transaction is a bundle,
+    /// this is the list of hashes of the transactions that reverted. If the
+    /// transaction is not a bundle, this is `None`.
+    pub reverted_hashes: Option<Vec<B256>>,
```
karim-agha, line 25: opinionated nit: should we just treat an empty vec as nothing? This might add an unnecessary `Option` check burden on the code user.

karim-agha (approved):
Looks good to me with minor opinionated nits that you can feel free to ignore.

crates/op-rbuilder/src/builders/context.rs
ferranbt: The transactions are never dropped from the pool unless we call the `remove_transaction` method. But, in this case we would have to mark it as invalid so that we do not consider its childs either like we do with the rest of the code.

PR #127, opened 2025-06-05
Update reverting transaction hashes JSON

trianglesphere:
## 📝 Summary

This renames the reverting_transaction_hashes field's JSON key to be `revertingTxHashes`. In addition, this makes the field optional in the JSON.

This builds but I am not able to get the tests to run locally.


ferranbt:
Fixed in #126. The problem is that if you use an Option in `set_reverted_hashes` during block building it cannot differentiate between a normal transaction and a bundle without reverting hashes and then it will include the bundle on the block instead of skipping it.

PR #136, opened 2025-06-06
Put the integration tests in their own feature

krl:
This is so that unit tests can be run without the whole development setup being there yet.

The feature is enabled by default, but unit tests can now be run separately.

### DIFF
--- a/crates/op-rbuilder/src/tests/vanilla/smoke.rs
+++ b/crates/op-rbuilder/src/tests/vanilla/smoke.rs
@@ -4,6 +4,7 @@ use std::collections::HashSet;
 
 /// This is a smoke test that ensures that transactions are included in blocks
 /// and that the block generator is functioning correctly.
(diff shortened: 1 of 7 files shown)


crates/op-rbuilder/src/tests/vanilla/smoke.rs
```diff
@@ -69,6 +70,7 @@ async fn chain_produces_blocks() -> eyre::Result<()> {
 /// Ensures that payloads are generated correctly even when the builder is busy
 /// with other requests, such as fcu or getPayload.
 #[tokio::test]
+#[cfg(feature = "integration_tests")]
```
karim-agha, line 73: I'd recommend putting those conditional compilation attributes at the module inclusion level rather than decorating every test case.

ferranbt:
Closing this as we now run everything on unit tests without any pre-setups.

PR #138, opened 2025-06-09
Split op-rbuilder in lib and main

ferranbt:
## 📝 Summary

Before, in the src folder we had both a `main.rs` function and an import `lib.rs` one. This setup created problems because methods exported in lib were not available for main since I believe the execution contexts were different.

This PR changes that and splits the op-rbuilder in two components, a library (under `src`) and the op-rbuilder main binary (under `bin`) which consumes the library. Other than the main split in functionality this PR does not introduce any other ergonomics about how to structure better the library to be consumed by the main binary. I will leave that to followup PRs that consume this functionality.

### DIFF
--- a/crates/op-rbuilder/src/launcher.rs
+++ b/crates/op-rbuilder/src/launcher.rs
@@ -2,8 +2,7 @@ use eyre::Result;
 
 use crate::{
     args::*,
-    builders::{BuilderConfig, PayloadBuilder},
-    builders::{BuilderMode, FlashblocksBuilder, StandardBuilder},
+    builders::{BuilderConfig, BuilderMode, FlashblocksBuilder, PayloadBuilder, StandardBuilder},
     metrics::VERSION,
     monitor_tx_pool::monitor_tx_pool,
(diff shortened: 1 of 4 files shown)


crates/op-rbuilder/src/launcher.rs
```diff
@@ -29,16 +22,14 @@ use reth_optimism_node::{
...
 };
 use reth_transaction_pool::TransactionPool;
-use revert_protection::{EthApiExtServer, EthApiOverrideServer, RevertProtectionExt};
 use std::{marker::PhantomData, sync::Arc};
-use tx::FBPooledTransaction;
 
 // Prefer jemalloc for performance reasons.
 #[cfg(all(feature = "jemalloc", unix))]
 #[global_allocator]
 static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
```
SozinM, line 30: We should move this to main, because lib shouldn't set global allocator

SozinM (approved):

Issue #35, opened 2025-05-13
Do not require building op-rbuilder to run the integration tests


ferranbt:
Right now if you want to run integration tests you have to create a binary of op-rbuilder. The framework will spawn the binary process on the background alongside other processes (i.e [L2P-497](https://linear.app/flashbots/issue/L2P-497/use-docker-containers-to-spawn-services)). This is a bit cumbersome because if you might forget about building the binary again after your changes, and it takes some time to figure out why the behaviour of the test is the same.

Instead, we could do like we do in Rollup-boost https://github.com/flashbots/rollup-boost/blob/main/tests/common/services/rollup_boost.rs#L58 which is running the op-rbuilder as an internal process.

PR #139, opened 2025-06-09
In process builder

ferranbt:
## 📝 Summary

Builds on top of #138 

This PR is an alternative design to solve #35. Major changes are:
- Refactored the main module (using #138) to introduce a `launch` function that can be use to spawn the block builder.
- Modifies the `TestHarness` to run the previous `launch` function. It creates a custom tracing subscriber to still output the logs to a log file.


crates/op-rbuilder/src/launcher.rs
```diff
@@ -89,17 +78,20 @@ where
     }
 }
 
-impl<B> Launcher<OpChainSpecParser, OpRbuilderArgs> for BuilderLauncher<B>
+pub struct NodeContext {
+    _node: Box<dyn Any>,
```
ferranbt, line 82: This is done because if the node reference is dropped the node stops + the trait for this reference is pretty complex to write.

ferranbt: I never heard that before :) Open to alternatives if you feel that is better

PR #141, opened 2025-06-10
Add minTimestamp and maxTimestamp as optional fields to bundle

akundaz:
## 📝 Summary
Add the `minTimestamp` and `maxTimestamp` fields to the bundle params.

## 💡 Motivation and Context

On Ethereum Mainnet we have two fields minTimestamp and maxTimestamp which are not on our L2 bundles. These fields were originally introduced on Ethereum L1 because PoW meant non-consistent block times, but they were then kept after The Merge because of missed slots. While we don’t want to use these features from a protocol perspective because we should not be trusting the wall clock on builder nodes, some searchers may still want this feature so we should allow for and simply add an asterisk that they are subject to builder node clock.

---

* [x] Run `make lint`
* [ ] Run `make test`


ferranbt:
You also have to modify the block building step to take into account the timestamp https://github.com/flashbots/op-rbuilder/blob/main/crates/op-rbuilder/src/builders/context.rs#L375

ferranbt (approved):
LGTM! One question would be, what if both block number and timestamp are set? What is the expected behaviour?

akundaz:
> LGTM! One question would be, what if both block number and timestamp are set? What is the expected behaviour?

They're basically separate filters, so a transaction needs to satisfy both in order to get included.

PR #145, opened 2025-06-11
Run the vanilla tests using both the flashblocks builder and the vanilla builder

karim-agha:
## 📝 Summary

This PR closes: 
- https://github.com/flashbots/op-rbuilder/issues/68

and builds on top of:
- https://github.com/flashbots/op-rbuilder/pull/132

## Test coverage

This change doubles our test coverage and reuses standard builder tests to cover flashbots builder as well.

## Automatic unit tests generation for both standard and flashblocks builders. 

When writing unit or integration tests instead of using `#[tokio::test] async test_name()` use the `rb_test` macro and let the `LocalInstance` be injected by the macro by the test infrastructure. In this mode a sample unit test would look as follows:

```rust
/// This test ensures that the transaction size limit is respected.
/// We will set limit to 1 byte and see that the builder will not include any transactions.
#[rb_test]
async fn data_availability_tx_size_limit(rbuilder: LocalInstance) -> eyre::Result<()> {
    let driver = rbuilder.driver().await?;

    // Set (max_tx_da_size, max_block_da_size), with this case block won't fit any transaction
    let call = rbuilder
        .provider()
        .await?
        .raw_request::<(i32, i32), bool>("miner_setMaxDASize".into(), (1, 0))
        .await?;
    assert!(call, "miner_setMaxDASize should be executed successfully");

    let unfit_tx = driver
        .transaction()
        .random_valid_transfer()
        .with_max_priority_fee_per_gas(50)
        .send()
        .await?;
    let block = driver.build_new_block().await?;

    // tx should not be included because we set the tx_size_limit to 1
    assert!(
        !block.includes(unfit_tx.tx_hash()),
        "transaction should not be included in the block"
    );

    Ok(())
}
```

then when you run `cargo test` it will automatically generate both variants of this test:

```
$ cargo test data_availability_tx_size_limit
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.29s
     Running unittests src/lib.rs (target/debug/deps/op_rbuilder-7194e393c77c3c5d)

running 2 tests
test tests::data_availability::data_availability_tx_size_limit_standard ... ok
test tests::data_availability::data_availability_tx_size_limit_flashblocks ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 33 filtered out; finished in 3.60s

     Running unittests src/main.rs (target/debug/deps/op_rbuilder-df17a41dd57b9b91)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 7 filtered out; finished in 0.00s

```

This macro can be configured to either run in only flashbots mode, only standard mode, use custom `RBuilderArgs` use custom `NodeConfig`, etc. Examples of using this macro are in the `tests` directory.

There is also a new macro called `if_flashblocks!` that allows you to specify a block of code that runs only under the flashblocks variant of a test. Example:

```rust
#[rb_test]
async fn test_name(rbuilder: LocalInstance) -> eyre::Result<()> {
  println!("This message will appear in both variants of the test");

  if_flashblocks! {
    println!("This message will appear only in flashblocks version of the test");
  }
}
```

Use this macro if the differences between variants are minimal and don't merit writing two versions of the test, otherwise use `#[rb_test(standard)]` and `#[rb_test(flashbots)]` to narrow down the generation to one builder.

---

* [X] Run `make lint`
* [X] Run `make test`
* [X] Added tests (if applicable)


crates/op-rbuilder/src/tests/framework/macros/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "macros"
```
krl, line 2: nitpick: maybe a less general name? something like "framework_test_macros"

krl (approved):
One minor thing, otherwise this looks very good!

PR #164, opened 2025-06-17
Make state root calculation configurable on Flashblocks builder

haardikk21:
## 📝 Summary

related to https://github.com/flashbots/rollup-boost/pull/335

Makes state root computation configurable in the builder, setting it to just be the zero value if not enabled. Used in conjunction with linked rollup-boost modifications to have the L2 client compute the state root instead.

## 💡 Motivation and Context

The context here is that state root computation per flashblock is extremely slow, and there are concerns if we can launch on Base Mainnet and still produce flashblocks every 200ms with that. By disabling that computation in the builder, and instead having rollup boost use the L2 Client to recompute the state root, we get around that problem by introducing a minor overhead in the getPayload request to rollup boost.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/flashblocks/payload.rs
+++ b/crates/op-rbuilder/src/builders/flashblocks/payload.rs
@@ -255,7 +255,12 @@ where
             .sequencer_tx_duration
             .record(sequencer_tx_start_time.elapsed());
 
-        let (payload, fb_payload, mut bundle_state) = build_block(db, &ctx, &mut info)?;
+        let (payload, fb_payload, mut bundle_state) = build_block(
+            db,
+            &ctx,
+            &mut info,
+            self.config.specific.calculate_state_root,
+        )?;
 
         best_payload.set(payload.clone());
         self.ws_pub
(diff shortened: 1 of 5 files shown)


ferranbt (approved):

haardikk21:
> @protolambda has an idea to replace the state root with a "flashhash" which is just a hash of the transactions and the previous flashblock flashhash to preserve ordering integrity, I wonder if we could fit that into this effort

what does this solve that we don't already get from flashblock index and cumulative txns/receipts list?

crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -684,26 +695,34 @@ where
...
-            })?
-    };
-    ctx.metrics
-        .state_root_calculation_duration
-        .record(state_root_start_time.elapsed());
+    // calculate the state root
+    let mut state_root = B256::ZERO;
+
+    if calculate_state_root {
+        let state_root_start_time = Instant::now();
```
cody-wang-cb, line 702: probably a nit but should we still emit this metric so we know for sure it's 0 rather than missing metric?

haardikk21: fair

avalonche:
@haardikk21 good to merge once rebase conflicts are resolved

PR #205, opened 2025-07-16
Add gauge metrics for block building steps

akundaz:
## 📝 Summary
Add gauge metrics for many metrics that happen once during block building. Also fix a few of the flashblocks metrics that were using the same payload metrics that normal block building used.

## 💡 Motivation and Context

Many of the current metrics that are recorded once per block are tracked with histograms, but it would be more useful to use gauges for them to get more accurate metrics for them. This is a somewhat common pattern, to use both a histogram and a gauge to track something, as seen in https://github.com/paradigmxyz/reth/pull/10728.

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/src/builders/context.rs
```diff
@@ -517,21 +517,37 @@ impl OpPayloadBuilderCtx {
...
+            .set(num_txs_simulated_success as f64);
         self.metrics
             .payload_num_tx_simulated_fail
             .record(num_txs_simulated_fail as f64);
+        self.metrics
+            .payload_num_tx_simulated_fail_gauge
+            .set(num_txs_simulated_fail as f64);
         self.metrics
             .bundles_reverted
             .record(num_bundles_reverted as f64);
```
avalonche, line 553: there are a lot of metrics getting recorded here, consider moving to a separate function

akundaz:
> would be great if there was a metrics test

Yeah, I'll look into this and try it out in a separate PR

Issue #197, opened 2025-07-10
feat: implement `gauge` metrics for key feature flags


0x416e746f6e:
it would be good for monitoring purposes for `op-rbuilder` to export `gauge` (values `0` or `1`) metrics for when this or that particular feature flag is enabled or not.  for example:

- `feat_revert_protection` => `1` when `--builder.enable-revert-protection` is in force, `0` otherwise
- `feat_flashblocks` => `1` when `--flashblocks.enabled` is in force, `0` otherwise
- `feat_flashtestations` => `1` for `--flashtestations.enabled`, `0` otherwise
- and so on...

0x416e746f6e:
> flag settings should be checked into git and current deployments should match the values in the repo

that's not possible in general case

akundaz:
Can you be more specific? This is a pretty standard practice. The only caveat I can think of is for secrets which are managed separately

PR #207, opened 2025-07-17
gauge metrics to inspect flag settings

akundaz:
## 📝 Summary
Add prometheus metrics to track if the following features are set:
* flashblocks
* flashtestations
* revert protection

## 💡 Motivation and Context

Resolves https://github.com/flashbots/op-rbuilder/issues/197

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/src/metrics.rs
```diff
@@ -155,6 +157,35 @@ impl OpRBuilderMetrics {
...
+/// Gauges to check whether particular flags are enabled or not
+#[derive(Metrics, Clone)]
+#[metrics(scope = "op_rbuilder_flags")]
+pub struct FlagMetrics {
+    /// Whether flashblocks.enabled is set or not
+    pub flashblocks_enabled: Gauge,
+    /// Whether flashtestations.enabled is set or not
+    pub flashtestations_enabled: Gauge,
+    /// Whether builder.enable-revert-protection is set or not
+    pub enable_revert_protection: Gauge,
```
0x416e746f6e, line 169: **suggestion:**

```suggestion
    /// Whether flashblocks.enabled is set or not
    pub feat_flashblocks_enabled: Gauge,
    /// Whether flashtestations.enabled is set or not
    pub feat_flashtestations_enabled: Gauge,
    /// Whether builder.enable-revert-protection is set or not
    pub feat_revert_protection_enabled: Gauge,
```

^-- this will keep them together and make it easier to search for them in metrics explorer.

akundaz: They will already all be together because I set this line above
```
#[metrics(scope = "op_rbuilder_flags")]
```
so we can search for them under this string

crates/op-rbuilder/src/metrics.rs
```diff
@@ -155,6 +157,35 @@ impl OpRBuilderMetrics {
     }
 }
 
+/// Gauges to check whether particular flags are enabled or not
+#[derive(Metrics, Clone)]
+#[metrics(scope = "op_rbuilder_flags")]
+pub struct FlagMetrics {
```
SozinM, line 163: We could do it even without structure, because we don't really using it
Just create a function that set all metrics with gauge! macros and metrics name

PR #214, opened 2025-07-24
feat: add transaction gas limit

### LINKED ISSUE #202: Add transaction gas limits
We want to filter transactions that use too much gas and discard them from being included in the built block. This is to prevent bad actors from degrading block building performance by taking up excessive resources. It should be configurable as a cli parameter.

varun-doshi:
## 📝 Summary

Adds a cli parameter named `max_gas_per_txn` to signiify if a transaction uses more gas than this value, it should not be added to the block being built.
Fixes #202 

* [x] Run `make lint`
* [x] Run `make test`


crates/op-rbuilder/src/builders/context.rs
```diff
@@ -566,6 +568,10 @@ impl<ExtraCtx: Debug + Default> OpPayloadBuilderCtx<ExtraCtx> {
 
                 // Add gas used by the transaction to cumulative gas used, before creating the receipt
                 let gas_used = result.gas_used();
+                if gas_used > self.max_gas_per_txn {
+                    info!("builder txn took execessive gas. not included in block");
+                    return Ok(());
```
SozinM, line 573: this is builder transaction, not needed here

SozinM:
Also this will need a test, looks good at this point
And waiting for @akundaz to review

varun-doshi:
I've added a utility function `random_big_transaction` which uses a random function data which is big. I can change this if needed.

- Tests added
- max_gas_per_txn is now optional

crates/op-rbuilder/src/tests/smoke.rs
```diff
@@ -189,3 +193,47 @@ async fn test_no_tx_pool(rbuilder: LocalInstance) -> eyre::Result<()> {
...
+})]
+async fn chain_produces_big_txs(rbuilder: LocalInstance) -> eyre::Result<()> {
+    let driver = rbuilder.driver().await?;
+
+    #[cfg(target_os = "linux")]
+    let driver = driver
+        .with_validation_node(crate::tests::ExternalNode::reth().await?)
+        .await?;
+
+    let count = rand::random_range(1..8);
```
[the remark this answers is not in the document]
varun-doshi: Well I just used this and removed the outer loop...I can add that back again

https://github.com/flashbots/op-rbuilder/blob/8909ca1d3d3c6e1ca5b2e7c58a582bc2e5d10d49/crates/op-rbuilder/src/tests/smoke.rs#L57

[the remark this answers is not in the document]
akundaz: I think a test that sends a tx using an amount of gas below the threshold and one above and then checking that the tx that uses too much gas is excluded should be sufficient

SozinM (changes requested):

crates/op-rbuilder/src/tests/smoke.rs
```diff
@@ -189,3 +193,70 @@ async fn test_no_tx_pool(rbuilder: LocalInstance) -> eyre::Result<()> {
...
+
+    // insert valid txn under limit
+    let tx = driver
+        .create_transaction()
+        .random_valid_transfer()
+        .send()
+        .await
+        .expect("Failed to send transaction");
+
+    // insert txn with gas usage above limit
```
akundaz, line 217: add an assertion to verify that this tx isn't in the block

crates/op-rbuilder/src/tests/smoke.rs
```diff
@@ -189,3 +193,70 @@ async fn test_no_tx_pool(rbuilder: LocalInstance) -> eyre::Result<()> {
...
+})]
+async fn chain_produces_big_tx_without_gas_limit(rbuilder: LocalInstance) -> eyre::Result<()> {
+    let driver = rbuilder.driver().await?;
+
+    #[cfg(target_os = "linux")]
+    let driver = driver
+        .with_validation_node(crate::tests::ExternalNode::reth().await?)
+        .await?;
+
+    // insert txn with gas usage but there is no limit
```
akundaz, line 248: similarly assert that this tx is included in the block

akundaz (changes requested):

akundaz (approved):

SozinM (approved):

SozinM (approved):

PR #226, opened 2025-08-05
fix: correctly calculate flashblock byte size

akundaz:
## 📝 Summary
Use the flashblock payload to correctly record the byte size

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/flashblocks/payload.rs
+++ b/crates/op-rbuilder/src/builders/flashblocks/payload.rs
@@ -2,7 +2,10 @@ use super::{config::FlashblocksConfig, wspub::WebSocketPublisher};
 use crate::{
     builders::{
         context::{estimate_gas_for_builder_tx, OpPayloadBuilderCtx},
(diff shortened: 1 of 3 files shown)


crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -535,7 +538,7 @@ where
                                 .record(flashblock_build_start_time.elapsed());
                             ctx.metrics
                                 .flashblock_byte_size_histogram
-                                .record(new_payload.block().size() as f64);
+                                .record(calculate_flashblock_byte_size(&fb_payload) as f64);
```
avalonche, line 541: there's no fb_payload.size() function? I think there should be an inbuilt function to calculate the byte size

SozinM: size would only calculate stack side of the structure

SozinM:
Found better way to solve it, will close this one

PR #227, opened 2025-08-06
fix: record num txs built with flashblocks enabled

akundaz:
## 📝 Summary
Change the location where we record the metrics for the number of txs in a block

## 💡 Motivation and Context
This wasn't being recorded correctly for flashblocks

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -300,13 +300,6 @@ where
             payload_id = fb_payload.payload_id.to_string(),
         );
 
-        ctx.metrics
```
SozinM, line 303: why have we moved those?

SozinM (approved):

PR #239, opened 2025-08-15
fix(flashblocks): gracefull cancellation on payload build failure

### LINKED ISSUE #209: Error logging on payload build failure and graceful cancellation token handling
If the payload building fails the flashblock_cancel_token_tx gets dropped which causes a send error when trying to send a child cancellation token in flashblocks. There is an error log indicating the channel has closed which is not helpful is diagnosing payload build failures. 
```
if let Some(Err(err)) = cancelled {
    error!(target: "payload_builder", "Timer task encountered error: {err}");
}
```

julio4:
## 📝 Summary

Close #209

Within a payload building job in flashblocks mode, we sequentially build flashblocks and a timer task coordinate timing of building these flashblocks. When a flashblock building fails, the whole payload building job fails and we need to gracefully shutdown all tasks related to this job. This is implemented by cancelling the payload building job cancellation token on flashblock building error, which correctly signal to the associated timer task to shutdown itself.

- Also fix channel variable names

## 💡 Motivation and Context

The channel between timer task to coordinate flashblock building was not gracefully shutdown when payload building fail, causing unwanted error and logs.

---

* [x] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/flashblocks/payload.rs
+++ b/crates/op-rbuilder/src/builders/flashblocks/payload.rs
@@ -369,11 +369,11 @@ where
         }
 
         // This channel coordinates flashblock building
-        let (fb_cancel_token_rx, mut fb_cancel_token_tx) =
+        let (fb_cancel_token_tx, mut fb_cancel_token_rx) =
             mpsc::channel((self.config.flashblocks_per_block() + 1) as usize);
         self.spawn_timer_task(
             block_cancel.clone(),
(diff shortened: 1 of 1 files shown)


julio4:
> I think with this we still won't get insight into the error we got

We do log the error [here](https://github.com/flashbots/op-rbuilder/pull/239/files#diff-20125ce86c70cfd520b09bc61e2032873f80249afe0027ee76dfbfcb503e7fb7L501)

Issue #203, opened 2025-07-15
Add per-address gas limits


akundaz:
We want to throttle users that use up too gas to protect block building performance. Otherwise, we might end up spending too much time to build blocks or build less valuable blocks. 

With per-address gas limits, bad actors trying to degrade our block building performance by distributing resource usage across multiple transactions would need to fund multiple addresses, raising the cost of such an attack.

The exact details of how to implement a per-user/address limit is unspecified.

PR #253, opened 2025-08-29
feat: address gas limiter

akundaz:
## 📝 Summary

Introduces a rate limiting mechanism for including transactions based on gas usage per address. 

## 💡 Motivation and Context

This is part of getting some basic DoS protection in the builder. With this limit it will be more expensive for attackers to compromise builder performance.

Resolves https://github.com/flashbots/op-rbuilder/issues/203

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/tests/framework/utils.rs
+++ b/crates/op-rbuilder/src/tests/framework/utils.rs
@@ -34,6 +34,8 @@ impl TransactionBuilderExt for TransactionBuilder {
         self.with_create().with_input(hex!("60006000fd").into()) // PUSH1 0x00 PUSH1 0x00 REVERT
     }
 
+    // This transaction is big in the sense that it uses a lot of gas. The exact
+    // amount it uses is 86220 gas.
     fn random_big_transaction(self) -> Self {
         // PUSH13 0x63ffffffff60005260046000f3 PUSH1 0x00 MSTORE PUSH1 0x02 PUSH1 0x0d PUSH1 0x13 PUSH1 0x00 CREATE2
         self.with_create()
--- a/crates/op-rbuilder/src/tests/gas_limiter.rs
+++ b/crates/op-rbuilder/src/tests/gas_limiter.rs
@@ -1,0 +1,121 @@
+use crate::{
+    args::OpRbuilderArgs,
+    gas_limiter::args::GasLimiterArgs,
+    tests::{ChainDriverExt, LocalInstance, TransactionBuilderExt},
+};
+use macros::rb_test;
+use std::collections::HashSet;
+use tracing::info;
+
+/// Integration test for the gas limiter functionality.
+/// Tests that gas limits are properly enforced during actual block building
+/// and transaction execution.
+#[rb_test(args = OpRbuilderArgs {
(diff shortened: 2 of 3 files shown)


crates/op-rbuilder/src/gas_limiter/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+#[derive(Debug, Clone)]
+struct TokenBucket {
+    capacity: u64,
+    available: u64,
+}
+
+impl AddressGasLimiter {
+    pub fn try_new(config: GasLimiterArgs) -> Option<Self> {
```
SozinM, line 41: We could make it better, make it return proper type, not an option and envoke only if config enables it
Even better if you couod always create this dtruct but it wouod do limiting logic onle when enabled, that wouod simplify a lot of code parts

crates/op-rbuilder/src/gas_limiter/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+pub mod args;
+pub mod error;
+mod metrics;
+
+#[derive(Clone)]
+pub struct AddressGasLimiter {
+    config: GasLimiterArgs,
+    // We don't need an Arc<Mutex<_>> here, we can get away with RefCell, but
+    // the reth PayloadBuilder trait needs this to be Send + Sync
+    address_buckets: Arc<Mutex<HashMap<Address, TokenBucket>>>,
```
SozinM, line 20: Let's just use DashMap

crates/op-rbuilder/src/gas_limiter/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use alloy_primitives::Address;
+
+use crate::gas_limiter::{args::GasLimiterArgs, error::GasLimitError, metrics::GasLimiterMetrics};
+
+pub mod args;
+pub mod error;
+mod metrics;
+
+#[derive(Clone)]
```
SozinM, line 15: Debug too

crates/op-rbuilder/src/gas_limiter/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Should be called upon each new block. Refills buckets/Garbage collection
+    fn refresh_inner(&self) -> usize {
+        let mut address_buckets = self.address_buckets.lock().unwrap();
+        let active_addresses = address_buckets.len();
+
+        address_buckets.iter_mut().for_each(|(_, bucket)| {
+            bucket.available += self.config.refill_rate_per_block;
+        });
+        address_buckets.retain(|_, bucket| bucket.available <= bucket.capacity);
```
SozinM, line 101: I think it's super bad for performance, if we created the bucket - let's keep it, otherwise we would always allocate memory 
Would make sense to remove bucket only if it wasn't used for 100 blocks (for example)

crates/op-rbuilder/src/gas_limiter/mod.rs
akundaz: done

PR #256, opened 2025-09-10
Add support for flashblocks number contract builder tx

avalonche:
## 📝 Summary

Integrates with the flashblocks number contract

## 💡 Motivation and Context

Allow contracts onchain to fill orders based on the flashblock number. Also improves observability of when each Flashblock starts.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/builder_tx.rs
+++ b/crates/op-rbuilder/src/builders/builder_tx.rs
@@ -106,8 +106,12 @@ pub trait BuilderTransactions<ExtraCtx: Debug + Default = ()>: Debug {
                 top_of_block,
             )?;
             for builder_tx in builder_txs.iter() {
-                if invalid.contains(&builder_tx.signed_tx.signer()) {
-                    warn!(target: "payload_builder", tx_hash = ?builder_tx.signed_tx.tx_hash(), "builder signer invalid as previous builder tx reverted");
+                let signed_tx = match builder_tx.signed_tx.clone() {
+                    Some(tx) => tx,
+                    None => continue,
+                };
+                if invalid.contains(&signed_tx.signer()) {
+                    warn!(target: "payload_builder", tx_hash = ?signed_tx.tx_hash(), "builder signer invalid as previous builder tx reverted");
                     continue;
                 }
 
@@ -210,7 +214,7 @@ impl BuilderTxBase {
                 Ok(Some(BuilderTransactionCtx {
                     gas_used,
                     da_size,
-                    signed_tx,
+                    signed_tx: Some(signed_tx),
                 }))
             }
             None => Ok(None),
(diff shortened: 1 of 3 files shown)


crates/op-rbuilder/src/builders/builder_tx.rs
```diff
@@ -167,3 +172,100 @@ pub trait BuilderTransactions<ExtraCtx: Debug + Default = ()>: Debug {
         Ok(simulation_state)
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct BuilderTxBase {
```
noot, line 177:
```suggestion
pub(crate) struct BuilderTxBase {
```

unless it's supposed to be part of the public API?

crates/op-rbuilder/src/builders/flashblocks/builder_tx.rs
```diff
@@ -1,28 +1,96 @@
...
+        builder_tx::{BuilderTxBase, get_nonce},
+        context::OpPayloadBuilderCtx,
+        flashblocks::payload::FlashblocksExtraCtx,
     },
     flashtestations::service::FlashtestationsBuilderTx,
     primitives::reth::ExecutionInfo,
     tx_signer::Signer,
 };
 
+sol!(
```
noot, line 32: do we foresee this code changing? it would be good to have a link/specific commit referenced for this

crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -294,10 +294,20 @@ where
...
-            self.builder_tx
-                .add_builder_txs(&state_provider, &mut info, &ctx, &mut state)?
+            match self
+                .builder_tx
+                .add_builder_txs(&state_provider, &mut info, &ctx, &mut state)
+            {
+                Ok(builder_txs) => builder_txs,
+                Err(e) => {
+                    error!(target: "payload_builder", "Error adding builder txs to fallback block: {}", e);
+                    vec![]
```
noot, line 304: is it okay to not include the builder txs? what are the consequences of not including them?

crates/op-rbuilder/src/builders/flashblocks/service.rs
```diff
@@ -92,10 +95,25 @@ where
...
+                    flashtestations_builder_tx,
+                ),
+            )
+        } else {
+            self.spawn_payload_builder_service(
+                ctx,
+                pool,
+                FlashblocksBuilderTx::new(signer, flashtestations_builder_tx),
+            )
+        }
```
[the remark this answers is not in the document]
avalonche: does this work? I think it'll cause complier errors

crates/op-rbuilder/src/builders/flashblocks/builder_tx.rs
avalonche: its subject to change depending on whether the flashblocks contract will rely on the flashtestations contract for auth. but I can link the flashblock contract repo

crates/op-rbuilder/src/builders/flashblocks/payload.rs
avalonche: but I think we can fallback to the regular builder tx if the flashblocks contract call fails so chain monitor can still mark the flashblock as something we built

PR #260, opened 2025-09-12
refactor:  clean up and improve flashblocks `build_payload`

noot:
## 📝 Summary

- refactor flashblocks `build_payload` into separate functions for readability, specifically moved the code executed within the loop to build a flashblock to its own `build_next_flashblock` fn (didn't change any logic here)
- removed `spawn_timer_task` in favour of select statements with ticker inside `build_payload` for clarity
- no code logic was changed with this PR, just refactored

## 💡 Motivation and Context

dug into #209 (this PR also improves the logging) and saw opportunity for clean up

## ✅ I have completed the following steps:

* [x] Run `make lint`
* [x] Run `make test`
* [ ] Added tests (if applicable)

PR #268, opened 2025-09-18
Add flashtestations builder transaction

avalonche:
## 📝 Summary

- Replaces `--debug-url` with `--quote_provider` arg
- Adds the flashtestations block proof at the end of the block 

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`


crates/op-rbuilder/src/flashtestations/tx_manager.rs
```diff
@@ -1,50 +1,47 @@
...
+    #[error("signer error: {0}")]
+    SignerError(ecdsa::Error),
+}
 
-    struct BlockData {
-        bytes32 parentHash;
-        uint256 blockNumber;
-        uint256 timestamp;
-        bytes32[] transactionHashes;
+impl From<TransportError> for TxManagerError {
```
akundaz, line 33: Instead of implementing this trait, use the thiserror "from" macro:
```
pub enum TxManagerError {
    #[error("rpc error: {0}")]
    RpcError(#[from] TransportError),
```

You can do this with any thiserror enum variants with unique values

crates/op-rbuilder/src/flashtestations/builder_tx.rs
```diff
@@ -0,0 +1,593 @@
...
+                gas_used,
+                success: true,
+                state_changes: state,
+                revert_reason: None,
+                logs,
+            }),
+            ExecutionResult::Revert { output, .. } => {
+                let revert_reason = FlashtestationRegistryError::from(output);
+                Ok(TxSimulateResult {
+                    gas_used: 0,
```
akundaz, line 251: Did you mean to fill out the `gas_used` field in this (and below under `Halt`)? you should have access to that value in `ExecutionResult::Revert`

crates/op-rbuilder/src/flashtestations/mod.rs
```diff
@@ -1,4 +1,211 @@
...
+    InvalidQuoteLength(U256),
+    #[error("invalid signature")]
+    InvalidSignature(),
+    #[error("invalid nonce: expected {0}, provided {1}")]
+    InvalidNonce(U256, U256),
+    #[error("unknown revert: {0}")]
+    Unknown(String),
+}
+
+impl From<Bytes> for FlashtestationRegistryError {
```
akundaz, line 96: There's a lot of repetitive code here, can you try something along the lines of the example here https://github.com/alloy-rs/examples/blob/main/examples/contracts/examples/revert_decoding.rs to reduce the amount of code here?

avalonche: I don't see how it reduces the lines of code here, we still need to convert each sol error type to our custom error type

crates/op-rbuilder/src/builders/builder_tx.rs
```diff
@@ -276,11 +271,20 @@ impl BuilderTxBase {
...
+pub fn get_nonce(
     db: &mut State<impl Database>,
     address: Address,
 ) -> Result<u64, BuilderTransactionError> {
     db.load_cache_account(address)
         .map(|acc| acc.account_info().unwrap_or_default().nonce)
         .map_err(|_| BuilderTransactionError::AccountLoadFailed(address))
 }
+
+pub fn get_balance(
```
noot, line 283:
```suggestion
pub(crate) fn get_balance(
```

crates/op-rbuilder/src/flashtestations/builder_tx.rs
```diff
@@ -0,0 +1,581 @@
...
+            ctx.base_fee(),
+            ctx.chain_id(),
+            nonce,
+        )?;
+        let ResultAndState { result, state } = match evm.transact(&register_tx) {
+            Ok(res) => res,
+            Err(err) => {
+                if err.is_invalid_tx_err() {
+                    warn!(target: "flashtestations", %err, "register tee service tx failed");
+                    return Ok(TxSimulateResult::default());
```
noot, line 232: why is it okay to ignore the error here?

crates/op-rbuilder/src/flashtestations/builder_tx.rs
```diff
@@ -0,0 +1,581 @@
...
+            gas_used,
+            success,
+            state_changes,
+            revert_reason,
+            logs,
+        } = self.simulate_register_tee_service_tx(ctx, evm)?;
+        if success {
+            if !self.check_tee_address_registered_log(&logs, self.tee_service_signer.address) {
+                warn!(target: "flashtestations", "transaction did not emit TEEServiceRegistered log, FlashtestationRegistry contract address may be incorrect");
+                Ok((None, false))
```
noot, line 420: not propagating the error here seems like it could get messy, why not propagate the errors and handle them in the caller?

crates/op-rbuilder/src/flashtestations/builder_tx.rs
avalonche: they are not fatal errors should be handled separately (nonce, balance, gas issues)

crates/op-rbuilder/src/flashtestations/builder_tx.rs
```diff
@@ -0,0 +1,581 @@
...
+        evm.modify_cfg(|cfg| {
+            cfg.disable_balance_check = true;
+        });
+
+        let mut builder_txs = Vec::<BuilderTransactionCtx>::new();
+
+        if !self.registered.get().unwrap_or(&false) {
+            info!(target: "flashtestations", "tee service not registered yet, attempting to register");
+            builder_txs.extend(self.fund_tee_service_tx(ctx, &mut evm)?);
+            let (register_tx, _) = self.register_tee_service_tx(ctx, &mut evm)?;
```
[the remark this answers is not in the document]
avalonche: see https://github.com/flashbots/op-rbuilder/pull/268#discussion_r2371147174

crates/op-rbuilder/src/builders/builder_tx.rs
avalonche: flashtestations crate can not access this function, need to reexport

crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -636,14 +633,7 @@ where
 
         // We got block cancelled, we won't need anything from the block at this point
         // Caution: this assume that block cancel token only cancelled when new FCU is received
-        if block_cancel.is_cancelled() {
-            self.record_flashblocks_metrics(
```
akundaz, line 640: Why are these two calls to `record_flashblocks_metrics` deleted?

crates/op-rbuilder/src/builders/builder_tx.rs
```diff
@@ -30,7 +30,10 @@ use crate::{
 pub struct BuilderTransactionCtx {
     pub gas_used: u64,
     pub da_size: u64,
-    pub signed_tx: Option<Recovered<OpTransactionSigned>>,
+    pub signed_tx: Recovered<OpTransactionSigned>,
+    // whether the transaction should be a top of block or
+    // bottom of block transaction
+    pub is_top_of_block: bool,
```
akundaz, line 36: If we want to switch to bottom of every flashblock how difficult will it be to change this? Not sure if we want something like this configurable or not but it might be easier to start that way

crates/op-rbuilder/src/builders/flashblocks/payload.rs
avalonche: since the refactor in https://github.com/flashbots/op-rbuilder/pull/260 record_flashblocks_metrics is [called again ](https://github.com/flashbots/op-rbuilder/blob/bf944d7e721c8aeeaa67f050e5c6d255421283ab/crates/op-rbuilder/src/builders/flashblocks/payload.rs#L524-L532) after the return from this function

crates/op-rbuilder/src/builders/builder_tx.rs
avalonche: in construction of each transaction added to builder_tx just flip the boolean

avalonche: oh you mean configuring via cli / env? I'm not sure this is something we want to change very often since changes to the builder tx position breaks external integrations / observability. Also the same builder tx have different positions in the block depending on whether the flashblock contract call is enabled or not etc. and the complexity would easily cause misconfiguration.

PR #275, opened 2025-09-24
feat: implement p2p layer and broadcast flashblocks

noot:
## 📝 Summary

- implement p2p crate which implements a simple libp2p node
- p2p node has the ability to support various stream protocols and broadcasting to all peers on a protocol
- note: this is not gossip; messages will only be sent to directly-connected peers. we currently assume a very small builder network (<10 nodes) so direct connectivity is okay, but for larger networks, this will need to be extended.
- update flashblocks payload builder to broadcast newly built flashblock payloads over p2p
- implement payload handler to handle incoming flashblock payloads from peers
- NOTE: syncing and execution of flashblocks is not yet implemented; it will be added to the payload handler in a follow up PR.

future extensions:
- connection management/max peer count/peer persistence
- flashblock sync and execution
- test fallback building (if one builder fails in the middle of a block after building some flashblocks, can another builder take over flashblock building?) be out of scope for this PR, as the main goal of this PR is to ensure flashblock sync over p2p.

## 💡 Motivation and Context
see https://hackmd.io/@nZ-twauPRISEa6G9zg3XRw/H1fekrwolx

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/builder_tx.rs
+++ b/crates/op-rbuilder/src/builders/builder_tx.rs
@@ -4,10 +4,7 @@ use alloy_evm::Database;
 use alloy_op_evm::OpEvm;
 use alloy_primitives::{
     Address, B256, Bytes, TxKind, U256,
-    map::{
-        DefaultHashBuilder,
-        foldhash::{HashSet, HashSetExt},
-    },
+    map::{HashMap, HashSet},
 };
 use alloy_sol_types::{ContractError, Revert, SolCall, SolError, SolInterface};
 use core::fmt::Debug;
@@ -33,7 +30,6 @@ use revm::{
     inspector::NoOpInspector,
     state::Account,
 };
-use std::collections::HashMap;
 use tracing::{trace, warn};
 
 use crate::{
@@ -44,7 +40,7 @@ use crate::{
 pub struct SimulationSuccessResult<T: SolCall> {
     pub gas_used: u64,
     pub output: T::Return,
-    pub state_changes: HashMap<Address, Account, DefaultHashBuilder>,
+    pub state_changes: HashMap<Address, Account>,
 }
 
 #[derive(Debug, Clone)]
@@ -196,7 +192,7 @@ pub trait BuilderTransactions<ExtraCtx: Debug + Default = (), Extra: Debug + Def
                 .evm_config
                 .evm_with_env(&mut *db, builder_ctx.evm_env.clone());
 
-            let mut invalid: HashSet<Address> = HashSet::new();
+            let mut invalid = HashSet::new();
 
             for builder_tx in builder_txs.iter() {
                 if builder_tx.is_top_of_block != top_of_block {


crates/p2p/src/lib.rs
```diff
@@ -0,0 +1,480 @@
...
+        compat::FuturesAsyncReadCompatExt as _,
+    };
+
+    let codec = LinesCodec::new();
+    let mut reader = FramedRead::new(stream.compat(), codec);
+
+    loop {
+        match reader.next().await {
+            Some(Ok(str)) => {
+                let payload: M = serde_json::from_str(&str)
```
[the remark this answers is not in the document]
noot: i originally had it there, i was debating if the serialization format should be internal or external to the p2p layer. i'm going to refactor this to not be json anyways. could change the p2p layer back to just handle `Bytes` instead of `Message` if you prefer that

crates/op-rbuilder/src/builders/flashblocks/ctx.rs
```diff
@@ -0,0 +1,75 @@
...
+            chain_spec,
+            max_gas_per_txn: builder_config.max_gas_per_txn,
+        })
+    }
+
+    pub(super) fn evm_config(&self) -> &OpEvmConfig {
+        &self.evm_config
+    }
+
+    pub(super) fn into_op_payload_builder_ctx(
```
avalonche, line 51: why not <Into> for conversions between the types?

noot: i'm probably going to remove this function after cleanup

crates/p2p/src/lib.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+impl Default for NodeBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl NodeBuilder {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
noot: there is `default()` also, I like having both

crates/op-rbuilder/src/builders/flashblocks/p2p.rs
```diff
@@ -0,0 +1,60 @@
+use alloy_primitives::U256;
+use reth::{core::primitives::SealedBlock, payload::PayloadId};
+use reth_optimism_payload_builder::OpBuiltPayload as RethOpBuiltPayload;
+use reth_optimism_primitives::OpBlock;
+use serde::{Deserialize, Serialize};
+
+pub(super) const AGENT_VERSION: &str = "op-rbuilder/1.0.0";
+pub(super) const FLASHBLOCKS_STREAM_PROTOCOL: p2p::StreamProtocol =
+    p2p::StreamProtocol::new("/flashblocks/1.0.0");
```
avalonche: are the versions meant to be updated manually

crates/op-rbuilder/src/builders/flashblocks/payload_handler.rs
```diff
@@ -0,0 +1,62 @@
...
+        } = self;
+
+        tracing::info!("flashblocks payload handler started");
+
+        loop {
+            tokio::select! {
+                Some(payload) = built_rx.recv() => {
+                    let _  = payload_events_handle.send(Events::BuiltPayload(payload.clone()));
+                    // TODO: only broadcast if `!no_tx_pool`?
+                    // ignore error here; if p2p was disabled, the channel will be closed.
```
avalonche: hmm what is p2p is enabled and an error occurs here?
we'll still want to broadcast on no_tx_pool because it contains state changes around deposit transactions that other builders need

crates/op-rbuilder/src/tests/flashblocks.rs
```diff
@@ -29,6 +26,8 @@ const FLASHBLOCKS_NUMBER_ADDRESS: Address = address!("5fbdb2315678afecb367f032d9
         flashblocks_block_time: 200,
         flashblocks_leeway_time: 100,
         flashblocks_fixed: false,
+        flashblocks_calculate_state_root: true,
```
avalonche: think this is redundant if default is specified

crates/p2p/src/behaviour.rs
```diff
@@ -0,0 +1,103 @@
+use eyre::WrapErr as _;
+use libp2p::{
+    autonat, connection_limits, connection_limits::ConnectionLimits, identify, identity, mdns,
+    ping, swarm::NetworkBehaviour,
+};
+use std::{convert::Infallible, time::Duration};
+
+const DEFAULT_MAX_PEER_COUNT: u32 = 200;
```
avalonche: maybe this can be an op-rbuilder arg

crates/p2p/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+
+impl<M: Message + 'static> Node<M> {
+    /// Returns the multiaddresses that this node is listening on, with the peer ID included.
+    pub fn multiaddrs(&self) -> Vec<libp2p::Multiaddr> {
+        self.listen_addrs
+            .iter()
+            .map(|addr| {
+                addr.clone()
+                    .with_p2p(self.peer_id)
+                    .expect("can add peer ID to multiaddr")
```
avalonche: what are the error cases when peer id can't be added?

crates/op-rbuilder/src/flashtestations/builder_tx.rs
```diff
@@ -2,7 +2,7 @@ use alloy_consensus::TxEip1559;
 use alloy_eips::Encodable2718;
 use alloy_evm::Database;
 use alloy_op_evm::OpEvm;
-use alloy_primitives::{Address, B256, Bytes, TxKind, U256, keccak256, map::foldhash::HashMap};
+use alloy_primitives::{Address, B256, Bytes, TxKind, U256, keccak256, map::DefaultHashBuilder};
```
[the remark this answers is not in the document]
noot: i'm not sure exactly why but the previous version doesn't compile for me, even though i didn't make changes:
```
error[E0308]: mismatched types
   --> crates/op-rbuilder/src/builders/builder_tx.rs:373:36
    |
373 |                     state_changes: state,
    |                                    ^^^^^ expected `HashMap<Address, Account, RandomState>`, found `HashMap<Address, Account, DefaultHashBuilder>`
    |
    = note: expected struct `HashMap<_, _, alloy_primitives::map::foldhash::fast::RandomState>`
               found struct `HashMap<_, _, DefaultHashBuilder>`

error[E0308]: mismatched types
   --> crates/op-rbuilder/src/flashtestations/builder_tx.rs:232:29
    |
232 |         evm.db_mut().commit(state_changes);
    |                      ------ ^^^^^^^^^^^^^ expected `HashMap<Address, Account, DefaultHashBuilder>`, found `HashMap<Address, Account, RandomState>`
    |                      |
    |                      arguments to this method are incorrect
    |
    = note: expected struct `HashMap<_, _, DefaultHashBuilder>`
               found struct `HashMap<_, _, alloy_primitives::map::foldhash::fast::RandomState>`
note: method defined here
   --> /home/e/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/revm-database-interface-7.0.5/src/lib.rs:72:8
    |
72  |     fn commit(&mut self, changes: HashMap<Address, Account>);
    |        ^^^^^^

For more information about this error, try `rustc --explain E0308`.
error: could not compile `op-rbuilder` (lib) due to 2 previous errors
```

i'm still trying to debug this but this was a stopgap change just so i can have the PR up still

crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -134,6 +134,9 @@ pub(super) struct OpPayloadBuilder<Pool, Client, BuilderTx> {
     pub pool: Pool,
     /// Node client
     pub client: Client,
+    /// Sender for sending built payloads to [`PayloadHandler`],
+    /// which broadcasts outgoing payloads via p2p.
+    pub payload_tx: mpsc::Sender<OpBuiltPayload>,
```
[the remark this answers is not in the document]
noot: it's standard for channels to have `_tx` or `_rx` at the end, so i renamed the metrics, let me know if that's okay

noot:
>  should we make params like trusted peers to make sure that only out builder would be able to link and send events?

do you mean so that only trusted peers are able to form connections? that could definitely be added as a flag, i plan to do a follow-up with more connection management features. i would say it's out of scope for this PR as it's already large, and is its own feature

PR #276, opened 2025-09-25
Add remote quote provider arg for flashtestations

avalonche:
## 📝 Summary

* Replaces --debug-url with --quote_provider arg

## 💡 Motivation and Context

Allow the external tdx provider to fetch attestations

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/flashtestations/args.rs
+++ b/crates/op-rbuilder/src/flashtestations/args.rs
@@ -23,17 +23,24 @@ pub struct FlashtestationsArgs {
     )]
     pub debug: bool,
 
-    // Debug url for attestations
-    #[arg(long = "flashtestations.debug-url", env = "FLASHTESTATIONS_DEBUG_URL")]
-    pub debug_url: Option<String>,
+    // Debug static key for the tee key. DO NOT USE IN PRODUCTION
+    #[arg(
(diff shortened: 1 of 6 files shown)


crates/op-rbuilder/src/flashtestations/args.rs
```diff
@@ -23,17 +23,24 @@ pub struct FlashtestationsArgs {
...
-    // Debug url for attestations
-    #[arg(long = "flashtestations.debug-url", env = "FLASHTESTATIONS_DEBUG_URL")]
-    pub debug_url: Option<String>,
+    // Debug static key for the tee key
+    #[arg(
+        long = "flashtestations.debug-tee-key-seed",
+        env = "FLASHTESTATIONS_DEBUG_TEE_KEY_SEED",
+        default_value = "debug"
+    )]
+    pub debug_tee_key_seed: String,
```
SozinM, line 32: Should this be Option<String>?

avalonche: it has a default value and won't be used if debug not enabled so it won't be a required arg. Option<String> will be same logic by having .unwrap_else("debug")

SozinM (approved):

PR #277, opened 2025-09-28
Add flashblocks number integration tests

avalonche:
## 📝 Summary

Refactor of builder tx and adds integration tests for flashblocks and flashtestations

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`


crates/op-rbuilder/src/builders/builder_tx.rs
```diff
@@ -276,11 +283,20 @@ impl BuilderTxBase {
...
+pub(crate) fn get_nonce(
     db: &mut State<impl Database>,
     address: Address,
 ) -> Result<u64, BuilderTransactionError> {
     db.load_cache_account(address)
         .map(|acc| acc.account_info().unwrap_or_default().nonce)
         .map_err(|_| BuilderTransactionError::AccountLoadFailed(address))
 }
+
+pub(crate) fn log_exists(logs: &[Log], topic: &B256) -> bool {
```
akundaz, line 295: Remove this function and do something like `logs.iter().any(|log| log.topics().first() == Some(topic))`

crates/op-rbuilder/src/builders/flashblocks/builder_tx.rs
```diff
@@ -33,62 +33,34 @@ use crate::{
...
 pub(super) enum FlashblockNumberError {
-    #[error("non builder address: {0}")]
-    NonBuilderAddress(Address),
-    #[error("mismatched flashblock number: expected {0}, actual {1}")]
-    MismatchedFlashblockNumber(U256, U256),
-    #[error("unknown revert: {0}")]
-    Unknown(String),
+    #[error("flashblocks number contract tx reverted: {0:?}")]
+    Revert(IFlashblockNumber::IFlashblockNumberErrors),
+    #[error("contract may be invalid, mismatch in log emitted: expected {0:?} but got {1:?}")]
```
akundaz, line 56: This error message isn't correct, right? You're checking all the logs but when you construct this error you're recording the first log's topic `logs.first().and_then(|log| log.topics().first().copied())`. Maybe a "not found" type message would be better

avalonche: there is only 1 event emitted so the log topic will always be first. This is to prevent successful transactions to empty addresses which may not be a flashblock number contract

avalonche: of you mean out of the first topics of all logs emitted. yep we can log a list of the first topic of each log.

crates/op-rbuilder/src/tests/flashblocks.rs
```diff
@@ -456,3 +455,208 @@ async fn test_flashblocks_no_state_root_calculation(rbuilder: LocalInstance) ->
...
+
+    // Mine initialization
+    driver.build_new_block_with_current_timestamp(None).await?;
+    provider
+        .get_transaction_receipt(*init_tx.tx_hash())
+        .await?
+        .expect("init tx not mined");
+
+    // Create transactions for testing contract calls
+    let transactions_phase2 = create_flashblock_transactions(&driver, 1..5).await?;
```
akundaz, line 543: Can you either separate this into two tests or make it more clear what the difference between phases 1 and 2 are in this longer test?

akundaz (approved):

PR #280, opened 2025-09-30
feat: add configurable maximum flashblocks per block cap (default 10)

cody-wang-cb:
## 📝 Summary
Adds a configurable maximum limit for the number of flashblocks that can be produced per block, ensures the maximum cannot exceed 10.

## 💡 Motivation and Context
The flashblock count was previously determined purely by timing calculations (`block_time / flashblock_interval`), which doesn't prevent the count go over 10 when using flashblock_interval < 200 for a 2s block time chain.

  ## Changes
  - **CLI Flag**: Added `--flashblocks.max-per-block` with default value of 10
  - **Environment Variable**: `FLASHBLOCKS_MAX_PER_BLOCK` support
  - **Configuration**: Added `max_flashblocks_per_block` field to `FlashblocksConfig`
  - **Logic**: Applied cap in both fixed and dynamic timing modes:
    - Fixed mode: Cap applied in `flashblocks_per_block()` method
    - Dynamic mode: Cap applied at the end of timing calculations
  - **Tests**: Added comprehensive test coverage for both timing modes

  ## Implementation Details
  The cap is applied in two places:
  1. `FlashBlocksConfigExt::flashblocks_per_block()` - handles fixed timing mode
  2. End of `calculate_flashblocks()` - handles dynamic timing mode calculations

  This approach ensures consistent behavior across both modes without restructuring the existing timing logic.

  ## Testing

  - ✅ test_max_flashblocks_per_block_cap - tests fixed timing mode with cap of 3
  - ✅ test_max_flashblocks_dynamic_timing - tests dynamic timing mode with cap of 5
  - ✅ All existing flashblocks tests continue to pass

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -166,6 +166,17 @@ pub struct FlashblocksArgs {
         env = "FLASHBLOCK_NUMBER_CONTRACT_ADDRESS"
     )]
     pub flashblocks_number_contract_address: Option<Address>,
+
+    /// Maximum number of flashblocks per block
+    ///
+    /// This caps the maximum number of flashblocks that can be produced per block,
+    /// regardless of the calculated number based on timing.
+    #[arg(
+        long = "flashblocks.max-per-block",
+        env = "FLASHBLOCKS_MAX_PER_BLOCK",
+        default_value = "10"
+    )]
+    pub flashblocks_max_per_block: u64,
 }
 
 impl Default for FlashblocksArgs {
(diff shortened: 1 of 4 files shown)


avalonche:
what's the use case for this? It could easily lead to misconfiguration using flashblock_interval < 200 for a 2s block time and the default caps it at 10. Could the default be unlimited?

avalonche:
I think it's possible to add a "buffer" for each flashblock to tune the interval instead of capping a maximum flashblock number. Would that achieve the same result? This seems more like a patch solution to the unintended behaviour of another hacky solution and easily leads to misconfiguration.

Issue #284, opened 2025-10-02
flashtestations:  persist TEE in `/run` directory


0x416e746f6e:
With flashtestations, `op-rbuilder` generates ephemeral flashtestation key on start, that it then uses to generate TDX quote.

On every restart it's a new key, which means that if `op-rbuilder` was restarted for whatever reason, it's a new key and new attestation, even if the TD box is still the same.

Since restarts can occur automatically (e.g. due to failure, and subsequent systemd unit auto-restart), it's preferred to keep the original ephemeral key around for the whole lifetime of the TD.

For example, the key could be kept in `/run/flashtestation.key` file (with `0600` perms) that would be written of first start, and re-read on each consecutive restart.

PR #286, opened 2025-10-04
Flag to save tee key to local file

avalonche:
## 📝 Summary
Resolves https://github.com/flashbots/op-rbuilder/issues/284

## 💡 Motivation and Context

Adding arg to save the key in local fs between restarts. File is guaranteed to never leave the TEE by the image measurements that includes all the config of the builder

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/flashtestations/service.rs
+++ b/crates/op-rbuilder/src/flashtestations/service.rs
@@ -1,9 +1,9 @@
-use alloy_primitives::{Bytes, keccak256};
+use alloy_primitives::{B256, Bytes, keccak256};
 use reth_node_builder::BuilderContext;
 use reth_provider::StateProvider;
 use reth_revm::State;
 use revm::Database;
-use std::fmt::Debug;
+use std::{fmt::Debug, fs, os::unix::fs::PermissionsExt, path::Path};
 use tracing::{info, warn};
 
 use crate::{
@@ -12,7 +12,7 @@ use crate::{
     },
     primitives::reth::ExecutionInfo,
     traits::NodeBounds,
-    tx_signer::{Signer, generate_ethereum_keypair, generate_key_from_seed},
+    tx_signer::{Signer, generate_key_from_seed, generate_signer},
 };
 
 use super::{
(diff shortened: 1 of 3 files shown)


crates/op-rbuilder/src/flashtestations/service.rs
```diff
@@ -135,6 +130,67 @@ where
...
+        info!("Flashtestations debug mode enabled, generating debug key from seed");
+        return Ok(generate_key_from_seed(debug_seed));
+    }
+
+    let path = Path::new(key_path);
+
+    // Try to load existing key
+    if path.exists() {
+        info!("Loading TEE key from {}", key_path);
+        match fs::read_to_string(path) {
```
akundaz, line 145: Can you use `inspect_err` or `unwrap_` functions to clean up the error handling?

avalonche: I don't want errors related to reading the key to fail the flashtestations bootstrap. I believe `inspect_err` and `unwrap_` will return the result

akundaz: It's difficult to follow the logic when you have nested statements. I extracted this into a separate function like this
```
fn try_load_key_from_path(key_path: &str) -> Option<Signer> {
    let path = Path::new(key_path);
    if !path.exists() {
        return None;
    }

    info!("Loading TEE key from {}", key_path);

    let key_hex = fs::read_to_string(path)
        .inspect_err(|e| warn!("failed to read key file: {:?}", e))
        .ok()?
        .trim();

    let secret_bytes = B256::try_from(
        hex::decode(key_hex)
            .wrap_err("failed to decode tee key")?
            .as_slice(),
    )
    .inspect_err(|e| warn!("failed to parse stored key: {:?}", e))
    .ok()?;

    Signer::try_from_secret(secret_bytes)
        .inspect_err(|e| warn!("failed to create signer from stored key: {:?}", e))
        .ok()
}
```
Now it's quicker to connect each error message with the corresponding function. 

Why do you return an error when you fail to decode the tee key but create a new one if you fail to parse it? I feel like you should have the same behavior in these scenarios

crates/op-rbuilder/src/flashtestations/service.rs
```diff
@@ -135,6 +130,67 @@ where
...
+            }
+        }
+    }
+
+    // Generate new key
+    info!("Generating new ephemeral TEE key");
+    let signer = generate_signer();
+
+    // Save key to file with 0600 permissions
+    let key_hex = hex::encode(signer.secret.secret_bytes());
```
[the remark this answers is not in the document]
Ruteri: Note the `/run` directory is indeed in ram. Saving to disk is a bit tricky, since disks can always be tampered with (harder to do with memory).

PR #287, opened 2025-10-04
Add permit functions for flashblocks number contract

avalonche:
## 📝 Summary

permit calls to flashblocks number contract if flashtestations are enabled as well. default to regular flashblocks number contract if flashtestations are not enabled. 

`--flashblocks.flashblocks-number-contract-use-permit` flag to enable.

---

* [x] Run `make lint`
* [x] Run `make test`


crates/op-rbuilder/src/builders/flashblocks/builder_tx.rs
```diff
@@ -137,96 +137,199 @@ impl BuilderTransactions<FlashblocksExtraCtx, FlashblocksExecutionInfo> for Flas
...
             signer,
             flashblock_number_address,
+            use_permit,
             base_builder_tx,
             flashtestations_builder_tx,
         }
     }
 
-    fn estimate_flashblock_number_tx_gas(
+    fn current_flashblock_number(
```
SozinM, line 167: Let's add docs
Will this function just returns current flashblock number? If so why does it needs mutable ref to state?

avalonche: then the read only contract calls only need the call data

SozinM: If you want this to be not mutable -  you need to use `basic_ref` function (that implemented for `DatabaseRef`)

SozinM (approved):

crates/op-rbuilder/src/builders/flashblocks/builder_tx.rs
```diff
@@ -242,46 +280,35 @@ impl BuilderTransactions<FlashblocksExtraCtx, FlashblocksExecutionInfo>
...
+                cfg.disable_balance_check = true;
+                cfg.disable_block_gas_limit = true;
+            });
 
-                let nonce = get_nonce(evm.db_mut(), signer.address)?;
-
-                let tx = match self.estimate_flashblock_number_tx_gas(ctx, &mut evm, signer, nonce)
-                {
-                    Ok(gas_used) => {
-                        // Due to EIP-150, 63/64 of available gas is forwarded to external calls so need to add a buffer
```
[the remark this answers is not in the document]
avalonche: this buffer is already calculated in the [sign_tx](https://github.com/flashbots/op-rbuilder/blob/8b7e42ac8047654771eab37b3cdf543ad2ea3c45/crates/op-rbuilder/src/builders/builder_tx.rs#L294) function

PR #288, opened 2025-10-06
feat: implement flashblock sync over p2p

noot:
## 📝 Summary

- implement sync and execution of flashblocks received over p2p

instructions for local testing with builder-playground, 1 builder building and 1 builder syncing:

1. checkout modified builder-playground branch: `cd builder-playground && git checkout noot/multi-builder`
2. checkout and build this branch: `cd op-rbuilder && git checkout noot/p2p-sync && cargo build`
3. run builder-playground:
```
sudo rm -rf ~/.playground ~/.local/share/reth && go run main.go cook opstack --flashblocks --external-builder http://host.docker.internal:4444 --flashblocks-builder ws://host.docker.internal:1111 --base-overlay
```
4. check that the sync-only node has started: `docker logs devnet-op-rbuilder-sync-only-1 -f`
5. run builder that will build:
```
./target/debug/op-rbuilder node --chain $HOME/.playground/devnet/l2-genesis.json --http --http.port 2222 --authrpc.port 4444 --authrpc.jwtsecret $HOME/.playground/devnet/jwtsecret --port 30333 --disable-discovery --rollup.builder-secret-key 59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d --trusted-peers enode://79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8@127.0.0.1:30304 --authrpc.addr 0.0.0.0 --flashblocks.enabled --flashblocks.addr 172.17.0.1 --flashblocks.port 1111 --rollup.chain-block-time 2000 --flashblocks.p2p_enabled
```

you should see logs showing that the nodes connected and that flashblocks are syncing and successfully executing and being added to the canonical chain:
```
[...]

PR #297, opened 2025-10-16
refactor: clean up flashblocks context in payload builder

noot:
## 📝 Summary

- clean up of flashblocks context in payload builder; this was originally part of #288 but I pulled it into its own PR as it's its own refactor
- main cleanup is changing `build_payload` such that the used `FlashblocksExtraCtx` isn't mutable, but a new one is constructed per each flashblock

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builders/flashblocks/payload.rs
+++ b/crates/op-rbuilder/src/builders/flashblocks/payload.rs
@@ -92,6 +92,17 @@ pub struct FlashblocksExtraCtx {
     calculate_state_root: bool,
 }
 
+impl FlashblocksExtraCtx {
+    fn next(self, target_gas_for_batch: u64, target_da_for_batch: Option<u64>) -> Self {
+        Self {
+            flashblock_index: self.flashblock_index + 1,
+            target_gas_for_batch,
+            target_da_for_batch,
+            ..self
+        }
+    }
+}
+
 impl OpPayloadBuilderCtx<FlashblocksExtraCtx> {
     /// Returns the current flashblock index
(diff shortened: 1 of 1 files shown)


crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -493,33 +511,57 @@ where
             };
             let _entered = fb_span.enter();
 
+            if ctx.flashblock_index() > ctx.target_flashblock_count() {
```
SozinM: We need to check that it's okay. AFAIK if future is completed before we get engine_getPayload it won't return result to it

noot: i tested this by removing `--flashblocks.enabled` for rollup-boost in builder-playground and running with telemetry, everything appears to be the same as before, what should i be looking for specifically to confirm?

crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -275,35 +242,78 @@ where
...
+            config.attributes.payload_attributes.id.to_string(),
+        );
+
+        let timestamp = config.attributes.timestamp();
+        let calculate_state_root = self.config.specific.calculate_state_root;
+        let ctx = self
+            .get_op_payload_builder_ctx(
+                config.clone(),
+                block_cancel.clone(),
+                FlashblocksExtraCtx {
```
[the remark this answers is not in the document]
noot: since `Default` is implemented on `FlashblocksExtraCtx` I replaced the empty fields w `..Default::default()`

PR #315, opened 2025-11-07
Add workload id as metric to builder

avalonche:
📝 Summary

  Adds workload ID computation and metrics to the builder for TEE workload
   identification. The workload ID uniquely identifies a TEE workload based on its measurement registers
  and is now exposed as a Prometheus metric for monitoring and observability.

  Key changes:
  - Implements compute_workload_id function that parses Automata's serialized verifier output and computes
   a unique workload identifier
  - Exposes workload ID as a Prometheus gauge metric with the workload ID as a label

  💡 Motivation and Context

  When running the builder in a TEE, it's important to be able to identify and monitor which specific
  workload/measurement is running

  By exposing this as a metric, operators can:
  - Verify the correct TEE workload is running to the one approved and registered unchain
  - Monitor and alert on workload changes
  - Ensure consistency across builder instances

  The implementation follows the
  https://github.com/flashbots/flashtestations/tree/7cc7f68492fe672a823dd2dead649793aac1f216 and same workload id from solidity and rust verified.

  ---
  ✅ I have completed the following steps:

  - Run make lint
  - Run make test
  - Added tests (if applicable)

### DIFF
--- a/crates/op-rbuilder/src/flashtestations/attestation.rs
+++ b/crates/op-rbuilder/src/flashtestations/attestation.rs
@@ -16,6 +16,23 @@ const TD_TDATTRS_VE_DISABLED: u64 = 0x0000000010000000;
 const TD_TDATTRS_PKS: u64 = 0x0000000040000000;
 const TD_TDATTRS_KL: u64 = 0x0000000080000000;
 
+// TD10ReportBody field offsets
+// These offsets correspond to the Solidity parseRawReportBody implementation
+const OFFSET_TD_ATTRIBUTES: usize = 120;
+const OFFSET_XFAM: usize = 128;
+const OFFSET_MR_TD: usize = 136;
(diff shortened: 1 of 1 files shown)


crates/op-rbuilder/src/flashtestations/rpc.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use super::{
+    args::FlashtestationsArgs,
+    attestation::{AttestationConfig, get_attestation_provider},
+    service::load_or_generate_tee_key,
+};
+
+/// Admin API for op-rbuilder flashtestations
+#[rpc(server, namespace = "admin")]
+pub trait AdminApi {
```
SozinM: Should we add this api to launcher.rs?

akundaz (approved):

PR #325, opened 2025-11-19
Add blob gas used to flashblocks delta

avalonche:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12808,7 +12808,7 @@ dependencies = [
 [[package]]
 name = "rollup-boost"
 version = "0.1.0"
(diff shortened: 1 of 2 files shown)


crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -1155,6 +1155,10 @@ where
             transactions: new_transactions_encoded,
             withdrawals: ctx.withdrawals().cloned().unwrap_or_default().to_vec(),
             withdrawals_root: withdrawals_root.unwrap_or_default(),
+            blob_gas_used: info
```
SozinM: let (excess_blob_gas, blob_gas_used) = ctx.blob_fields(info);

PR #328, opened 2025-11-20
chore(op-alloy): Update to use op-alloy flashblock types

0x00101010:
## 💡 Motivation and Context

Update to use op-alloy flashblock types. For consistency across repos.

* Corresponding rollup-boost PR: https://github.com/flashbots/rollup-boost/pull/441
* Reth reference: https://github.com/paradigmxyz/reth/pull/19608

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -167,11 +167,11 @@ rollup-boost = { git = "https://github.com/flashbots/rollup-boost", tag = "v0.7.
 # optimism
 alloy-op-evm = { version = "0.23.0", default-features = false }
 alloy-op-hardforks = "0.4.4"
-op-alloy-rpc-types = { version = "0.22.0", default-features = false }
-op-alloy-rpc-types-engine = { version = "0.22.0", default-features = false }
-op-alloy-rpc-jsonrpsee = { version = "0.22.0", default-features = false }
-op-alloy-network = { version = "0.22.0", default-features = false }
-op-alloy-consensus = { version = "0.22.0", default-features = false }
+op-alloy-rpc-types = { version = "0.22.4", default-features = false }
+op-alloy-rpc-types-engine = { version = "0.22.4", default-features = false }
+op-alloy-rpc-jsonrpsee = { version = "0.22.4", default-features = false }
+op-alloy-network = { version = "0.22.4", default-features = false }
+op-alloy-consensus = { version = "0.22.4", default-features = false }
 op-alloy-flz = { version = "0.13.1", default-features = false }
 
 async-trait = { version = "0.1.83" }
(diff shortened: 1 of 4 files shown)


SozinM:
@0x00101010 There is websocket proxy that work with flashblock but do not import anything from reth/alloy 
https://github.com/flashbots/rollup-boost/blob/main/crates/websocket-proxy/Cargo.toml

Any infra tool that would need flashblock shouldn't import alloys

SozinM:
I'm up for unification, but i prefer it to be separate crate, without other deps, and esp without heavy deps

0x00101010:
> @0x00101010 There is websocket proxy that work with flashblock but do not import anything from reth/alloy https://github.com/flashbots/rollup-boost/blob/main/crates/websocket-proxy/Cargo.toml
> 
> Any infra tool that would need flashblock shouldn't import alloys

1. websocket-proxy only uses raw data, compress it and pass around, it doesn't need any of the detailed data structures?
2. re in another crate, still think importing op-alloy is the cleaner choice:
    a. FlashBlockMetadata contains reference to OpReceipt which is defined in op-alloy, > @0x00101010 There is websocket proxy that work with flashblock but do not import anything from reth/alloy https://github.com/flashbots/rollup-boost/blob/main/crates/websocket-proxy/Cargo.toml
> 
> Any infra tool that would need flashblock shouldn't import alloys

1. websocket-proxy only uses raw data, compress it and pass around, it doesn't need any of the detailed data structures?
2. re in another crate, still think importing op-alloy is preferable due to:
    a. FlashBlockMetadata contains reference to OpReceipt which is defined in op-alloy, > @0x00101010 There is websocket proxy that work with flashblock but do not import anything from reth/alloy https://github.com/flashbots/rollup-boost/blob/main/crates/websocket-proxy/Cargo.toml
> 
> Any infra tool that would need flashblock shouldn't import alloys

1. websocket-proxy only uses raw data, compress it and pass around, it doesn't need any of the detailed data structures?
2. re in another crate, still think importing op-alloy is preferable due to:
    a. FlashBlockMetadata contains reference to OpReceipt which is defined in op-alloy. If we create a new standalone crate, we either depend on op-alloy anyway or duplicate those types, which is unnecessary and error-prone.
    c. Having the primitives live in op-alloy means there’s a single canonical representation that other OP tooling can reuse. A separate primitives crate under Flashbots effectively creates a parallel dialect that third parties have to adapt to.

I totally understand the concern around slowing down iteration by putting this in op-alloy. In practice though, we have a safety valve: if we ever need a hot patch, we can fork the repo, make the change, and temporarily depend on that fork via git until an official release is cut. That lets us keep the types canonical in op-alloy without sacrificing our ability to move fast in edge cases.

PR #349, opened 2026-01-03
Implement TaskGuard RAII pattern and fix Block-STM livelock

meyer9:
## Summary

This PR implements two major improvements to the Block-STM parallel execution engine:

### 1. TaskGuard RAII Pattern
- Refactored `num_active_tasks` tracking to use a RAII guard pattern
- TaskGuard automatically increments counter on creation and decrements on drop
- Eliminates manual increment/decrement tracking that was error-prone
- Guards are created only when tasks are successfully acquired
- Guards can be reused when `finish_execution`/`finish_validation` return follow-up tasks

### 2. Fixed MVHashMap Livelock Bug
- **Root cause**: When converting writes to estimates during validation abort, old write entries weren't removed from the MVHashMap
- This caused subsequent reads to get `Estimate` values instead of actual `Write` values
- Transactions would infinitely retry when `add_dependency()` returned false (dependency already satisfied)
- **Fix**: Remove old write entries before updating `last_written_locations` in `convert_writes_to_estimates()`

### Additional Changes
- Removed debug logging added during troubleshooting
- Updated all scheduler methods to return `Option<(Task, TaskGuard)>` tuples
- Updated executor to destructure and pass guards through the task lifecycle
- Cleaned up debug code in context.rs

## Test plan
- [ ] Run existing Block-STM tests to ensure no regressions
- [ ] Verify `fee_priority_ordering_standard` test no longer hangs
- [ ] Test with parallel execution mode enabled
- [ ] Confirm no livelock issues under high transaction load

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/op-rbuilder/src/block_stm/executor.rs
+++ b/crates/op-rbuilder/src/block_stm/executor.rs
@@ -18,7 +18,7 @@ use revm::{
     state::{Account, EvmState},
 };
 use tokio_util::sync::CancellationToken;
-use tracing::{Span, debug, warn};
+use tracing::{Span, warn};
 
 use crate::{
     block_stm::{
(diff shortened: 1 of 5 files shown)


meyer9:
sorry, gh command created on the wrong base

PR #355, opened 2026-01-16
Update to reth 1.10.0

avalonche:
## 📝 Summary

Reth changed StateProvider to not have the Sync trait so refactored flashblocks payload building to not be asynchronous.

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`


SozinM (approved):

crates/op-rbuilder/src/builders/flashblocks/payload.rs
```diff
@@ -931,7 +892,7 @@ where
 
         debug!(
             target: "payload_builder",
-            message = message,
+            message = "Payload building complete, job cancelled or target flashblock count reached",
```
akundaz: Why lose the granularity from before? We might record metrics at different times

PR #445, opened 2026-03-15
fix: use cumulative prefix sets for incremental trie state root

avalonche:
Fix: track cumulative TriePrefixSetsMut across all flashblocks. Before building TrieInput, extend the current prefix sets with all prior flashblocks' prefix sets. This forces the walker to re-visit every path modified in earlier flashblocks.

The fix is also ~30% faster than the unfixed incremental path because descending into cached in-memory nodes is faster than DB cursor seeks for skipped branches.

Benchmarks (10 flashblocks, 50k accounts):
- Without cache:           ~2,200ms (baseline)
- Incremental (no fix):    ~844ms  (2.6x faster, incorrect on reverts)
- Incremental (with fix):  ~650ms  (3.4x faster, correct)


docs/TRIE_CACHE_BENCHMARK_REPORT.md
julio4: We can keep the bench but maybe not the result?
If we choose to keep the result, atleast move it in `crates/op-builder/benches/docs or results`.

julio4 (approved):

akundaz (changes requested):
The benchmark and tests run against replicated state root calculations instead of the actual production code path. Can you extract the code to run this calculation and and update the benchmark and tests to call that directly? Otherwise we can't really draw conclusions about real world performance from these results. And if the code drifts we should be able to catch with our tests and the benchmark.

crates/op-rbuilder/src/builder/mod.rs
```diff
@@ -19,6 +19,7 @@ mod p2p;
 mod payload;
 mod payload_handler;
 mod service;
+pub mod state_root;
```
akundaz: make this not `pub` to match the convention used by the other modules

akundaz (changes requested):
Just adding a few comments. I think once they are resolved and CI passes this will probably be good

PR #458, opened 2026-04-09
feat: continuous-build mode

julio4:
## 📝 Summary

Adds a continuous flashblock building mode (under `--flashblocks.continuous-build` flag) that pre-seals candidates between scheduler triggers, so the best block is ready for instant publish when the timer fires.


crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,808 @@
...
+
+            candidates_evaluated += 1;
+
+            match build_result {
+                Ok((new_payload, mut fb_payload)) => {
+                    fb_payload.index = flashblock_index;
+                    fb_payload.base = None;
+
+                    let best_gas_used = best.as_ref().map_or(0, |b| b.gas_used);
+                    let is_new_best =
```
avalonche, line 749: why "better" is cumulative_gas_used instead of priority fee tips?

crates/op-rbuilder/src/metrics.rs
```diff
@@ -181,6 +181,15 @@ pub struct OpRBuilderMetrics {
...
     pub backrun_transaction_processing_gauge: Gauge,
+    /// Duration the continuous builder ran before being interrupted/completing
+    pub continuous_build_duration: Histogram,
+    /// How stale the best candidate was when published (time since sealed)
+    /// A high stale value can either mean no improvement in new candidates or slow candidate building
+    pub candidate_staleness: Histogram,
+    /// Number of candidates evaluated per flashblock interval
+    pub continuous_candidates_evaluated: Histogram,
+    /// Number of candidates that improved the latest best candidate
+    pub continuous_candidates_improved: Histogram,
```
avalonche, line 192: on what metric is the candidate improved. Instead of number of candidates I think its useful to have a histogram of how much it "improved" instead

julio4: gas used, and yes agree good idea

crates/op-rbuilder/src/builder/continuous.rs
julio4: my intuition is that the candidate know how to order to i.e. maximize prio fee; but the outer candidate loop only try to maximize filling blocks. But there may be cases where best candidate is selected when we had a better one in terms of fees

crates/op-rbuilder/src/builder/payload.rs
```diff
@@ -54,11 +54,23 @@ use reth_tasks::Runtime;
...
+        Arc,
+        atomic::{AtomicU64, Ordering},
+    },
+    time::{Duration, Instant},
+};
 use tokio::sync::{mpsc, watch};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, info_span, metadata::Level, span, warn};
 
+#[path = "continuous.rs"]
```
akundaz, line 70: Can you add the continuous module to mod.rs and import types from there the same way we import from other modules?

crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,808 @@
+use super::*;
```
akundaz, line 1: We only use this glob import in test modules, please import the types for this module in usual way

crates/op-rbuilder/src/builder/service.rs
```diff
@@ -109,6 +117,24 @@ impl FlashblocksServiceBuilder {
 
         let metrics = Arc::new(OpRBuilderMetrics::default());
         let task_metrics = Arc::new(FlashblocksTaskMetrics::new());
+        let pool_change_epoch = Arc::new(AtomicU64::new(0));
```
akundaz, line 120: What is the purpose of this pool change atomic?

crates/op-rbuilder/src/builder/payload.rs
```diff
@@ -91,6 +103,56 @@ type NextFlashblockPoolTxCursor<'a, Pool> = FlashblockPoolTxCursor<
     >,
 >;
 
+/// Adjust batch gas/DA/uncompressed limits to reserve capacity for bottom-of-block builder txs.
+/// Returns the adjusted `max_uncompressed_block_size`.
+pub(super) fn reserve_builder_tx_budget(
```
akundaz, line 108: Can you spin this out into a separate PR? Maybe we can get it in quicker

crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,808 @@
...
+use crate::builder::cancellation::PayloadJobCancellation;
+use tokio::sync::{mpsc, oneshot, watch};
+use tracing::{debug, error, info, metadata::Level, span, warn};
+
+pub(super) struct ContinuousBuildState<Cache, Transition> {
+    pub(super) ctx: OpPayloadBuilderCtx,
```
akundaz, line 7: The fields of this struct aren't accessed outside this module, right? You can add a constructor and make the fields private

crates/op-rbuilder/src/builder/continuous.rs
avalonche: what if a tx comes in with high fees during flashblock block building, txs with high gas usage would be prioritized over the high fee tx no? I think we should always use priority fee as the metric as a better block

crates/op-rbuilder/src/builder/payload.rs
julio4: I would say continuous is a submodule of payload but let's do that for now.

crates/op-rbuilder/src/builder/payload.rs
julio4: #476

avalonche:
did you push your changes correctly? don't see code changes for comments

crates/op-rbuilder/src/builder/service.rs
julio4: It lets the continuous loop detect pool changes without awaiting pool events (i.e. avoiding async overhead in the hot path). It coalesces all pool events into a single 'pool changed' signal, which also lets the loop skip useless iterations (if the pool hasn't changed, we can't produce a better candidate)

crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,957 @@
...
+    Result<ContinuousBuildOutput<CacheState, Option<TransitionState>>, PayloadBuilderError>,
+>;
+
+/// Slot that holds the latest sealed candidate from the build task.
+/// Task writes on each improvement; main loop takes on trigger to publish
+/// without awaiting task completion.
+type SharedBest = Arc<std::sync::Mutex<Option<BestCandidate>>>;
+
+/// Returned by `publish_and_spawn_next` to tell the caller whether to continue or exit.
+enum FlashblockAction {
```
akundaz: Do you plan on extending this type? It looks like a Result type would work just as well

crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,957 @@
...
+            let current_epoch = self.pool_change_epoch().load(Ordering::Relaxed);
+            if best.is_some() && current_epoch == last_seen_pool_change_epoch {
+                std::thread::sleep(idle_backoff);
+                idle_backoff = (idle_backoff * 2).min(max_idle_backoff);
+                continue;
+            }
+            last_seen_pool_change_epoch = current_epoch;
+            idle_backoff = Duration::from_millis(1);
+
+            // Yield CPU briefly to avoid starving the tokio runtime
```
akundaz: Shouldn't you run most of this (creating a new candidate) in a blocking task and then await it in each loop iteration?

crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,957 @@
...
+            info!(
+                target: "payload_builder",
+                candidates_evaluated,
+                candidates_improved,
+                staleness_ms = staleness_ms.unwrap_or(0),
+                "Scheduler trigger fired, publishing best candidate"
+            );
+        }
+
+        if payload_cancel.is_resolved() || payload_cancel.is_new_fcu() {
```
akundaz: Are you not able to make this function async and race against this token?

crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,957 @@
...
+                gas_used = tracing::field::Empty,
+                candidates_evaluated = tracing::field::Empty,
+                candidates_improved = tracing::field::Empty,
+                staleness_ms = tracing::field::Empty,
+            )
+        };
+
+        state.ctx = state.ctx.with_cancel(new_fb_cancel);
+        let build_start = Instant::now();
+        let new_shared_best: SharedBest = Arc::new(std::sync::Mutex::new(None));
```
akundaz: You already create a `SharedBest` up in `run_continuous_flashblocks` so you should use that one

crates/op-rbuilder/src/builder/continuous.rs
```diff
@@ -0,0 +1,957 @@
...
+                block_number = state.ctx.block_number(),
+                tx_count = tracing::field::Empty,
+                gas_used = tracing::field::Empty,
+                candidates_evaluated = tracing::field::Empty,
+                candidates_improved = tracing::field::Empty,
+                staleness_ms = tracing::field::Empty,
+            )
+        };
+
+        let mut shared_best: SharedBest = Arc::new(std::sync::Mutex::new(None));
```
akundaz: `Arc<Mutex<..>>` gives you interior mutability so please don't also use the usual `mut`

julio4: No I want each flashblock interval task to owns its slot for its lifetime. The easiest way is to rebind to a new Arc<Mutex<.>>. build tasks have long blocking part, so a 'cancelled-but-not-yet-stopped' build task could write a stale candidate into the slot after the the new task, with this it would just be automatically dropped and discarded.

crates/op-rbuilder/src/builder/continuous.rs
julio4: Shape is similar but Exit variants is not an error and is expected, and we already have `Result<_, payloadBuilderError>` wrapped. But I can use control flow

crates/op-rbuilder/src/builder/continuous/interval.rs
```diff
@@ -0,0 +1,216 @@
...
+                            &interval.base_info,
+                            target_flashblocks,
+                            deps.span,
+                        );
+                        drop(interval.build_rx);
+                        return Ok(());
+                    }
+                },
+            }
+        }
```
akundaz, line 87: A comment on why `publish_and_spawn_next` may signal to break from this loop would be nice.

Also, the metrics recording looks the same in two of the branches, you could instead break from the loop and do that after instead of in the `select!` body

crates/op-rbuilder/src/builder/continuous/shared_best.rs
```diff
@@ -0,0 +1,194 @@
+use super::types::BestCandidate;
+use std::sync::{Arc, Mutex, MutexGuard};
+use tracing::warn;
+
+/// Slot that holds the current highest-fee sealed [`BestCandidate`] from the build task.
+/// Task writes on each improvement; main loop takes on trigger to publish
+/// without awaiting task completion.
+pub(super) type SharedBest = CandidateSlot<BestCandidate>;
```
akundaz: If you define `SharedBest` to be a proper struct instead of a type alias, you can rewrite the free functions in this file as methods on the struct and it will be a lot easier to read

crates/op-rbuilder/src/builder/continuous/candidate_loop.rs
```diff
@@ -0,0 +1,454 @@
...
+                    fb_state: empty_candidate_fb_state,
+                    tx_tracker: base_tx_tracker.clone(),
+                    result: (next_flashblock_state, new_payload, fb_payload),
+                    build_duration: empty_build_duration,
+                    gas_limiter_snapshot: ctx.address_gas_limiter.snapshot(),
+                    candidates_evaluated,
+                    candidates_improved: candidates_improved + 1,
+                };
+                store_shared_best(shared_best, candidate.clone());
+                best = Some(candidate);
```
akundaz: ⚠️ Why do you need to store the best candidate twice?

crates/op-rbuilder/src/builder/continuous/interval.rs
```diff
@@ -0,0 +1,216 @@
...
+                    Self::record_cancellation_reason(self.metrics(), deps.payload_cancel, deps.span);
+                    self.record_flashblocks_metrics(
+                        &interval.base_ctx,
+                        &interval.base_fb_state,
+                        &interval.base_info,
+                        target_flashblocks,
+                        deps.span,
+                    );
+                    return Ok(());
+                }
```
akundaz: `payload_cancel` is triggered when we receive the `getPayload` call, right? I figured we could publish the last flashblock then as well but here you're just exiting the loop. Is it not possible to do that? We would end up solving the race with ending block building and sending the last flashblock if possible

crates/op-rbuilder/src/builder/continuous/transition.rs
```diff
@@ -0,0 +1,133 @@
...
+        );
+    }
+
+    #[test]
+    fn fallback_no_candidate_records_interval_base() {
+        assert_eq!(
+            fallback_no_candidate_metrics_source(),
+            StopMetricsSource::IntervalBase
+        );
+    }
```
akundaz, line 199: I haven't looked through the tests in this PR, but please make sure they're useful. This one in particular doesn't add anything meaningful since `fallback_no_candidate_metrics_source` just returns this constant value.

crates/op-rbuilder/src/builder/continuous/publish.rs
```diff
@@ -0,0 +1,557 @@
...
+        {
+            let _enter = fb_span.enter();
+            info!(
+                target: "payload_builder",
+                event = event.as_str(),
+                candidates_evaluated,
+                candidates_improved,
+                "Continuous trigger handled"
+            );
+        }
```
akundaz: What's the point of this span? It seems like you can just add the fields to the log statement

akundaz (dismissed):

crates/op-rbuilder/src/builder/continuous/transition.rs
julio4: will remove this one but all the other tests here acts as truth table checks, this last one was for completeness.

crates/op-rbuilder/src/builder/continuous/publish.rs
julio4: This span traces the whole lifecycle of one flashblock building, which is very useful to visualize all timings. Candidates_evaluated/improved is slightly redundant with the continuous trigger logs, it was to have all info directly in the span without having to go find the exact log.

crates/op-rbuilder/src/builder/continuous/interval.rs
julio4: payload_cancel is triggered when one of these fire: getPayload received, payload deadline reached, new FCU received.

I assumed that last trigger should always arrive before getPayload, but I agree with you this would be a nice improvement to submit last flashblock here if it's not the case.

crates/op-rbuilder/src/builder/continuous/candidate_loop.rs
julio4: Hum I see why it can seems slightly confusing but `shared_best` is the cross-thread channel so that the main building loop (outside of the continuous loop task, similar to the `best_payload` channel) always have access to "latest best candidate" without having to race anything on what continuous loop is doing. Because of that `shared_best` can be taken at any time hence why it's cloned here, and `best` is the local build state.

akundaz: Has there been a gap between cancelling the candidate loop and it actually ending? I wonder how necessary it is to do this. It would be simpler if the candidate loop just returned the best candidate

crates/op-rbuilder/src/builder/continuous/types.rs
```diff
@@ -0,0 +1,93 @@
...
+/// fresh `Arc<Mutex<_>>` per interval so a cancelled-but-not-yet-stopped
+/// build task cannot overwrite the new task's slot.
+pub(super) struct FlashblockInterval {
+    pub(super) fb_span: tracing::Span,
+    pub(super) build_rx: BuildReceiver,
+    pub(super) candidate_slot: SharedBest,
+    pub(super) build_start: Instant,
+    pub(super) base_ctx: OpPayloadJobCtx,
+    pub(super) base_fb_state: FlashblocksState,
+    pub(super) base_info: ExecutionInfo,
```
akundaz, line 103: are these fields just used for metrics? if so can you find a better way of recording those metrics instead of carrying these clones around? It's not clear otherwise why you would need to duplicate this info from `BuildState`

crates/op-rbuilder/src/builder/continuous/interval.rs
akundaz: definitely, i think leveraging `PayloadJobCancellation` would be a good idea

julio4: Yes but let's do this as a follow up

crates/op-rbuilder/src/builder/continuous/candidate_loop.rs
akundaz: Also, have you looked into using a tokio watch channel to track the best candidate? It seems like a natural fit to this problem

julio4: watch channel is not the best fit here as there's a single consumer once and take ownership of the result, the channel adds a lot of useless mechanisms.

avalonche (approved):

PR #459, opened 2026-04-09
Upgrade reth to 2.0

avalonche:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/src/tests/framework/instance.rs
```diff
@@ -354,8 +354,7 @@ fn chain_spec() -> Arc<OpChainSpec> {
 }
 
 fn task_runtime() -> TaskRuntime {
```
akundaz, line 348: nit: seems like we can just remove this function now

akundaz (commented):
looks generally fine to me, let's just get ci passing and test it out in some env

julio4 (approved):

PR #464, opened 2026-04-14
feat: add flashblock publish timing metric

julio4:
New `op_rbuilder_flashblock_publish_timing_milliseconds` gauge metric that records when each flashblock is published relative to slot start. This enable builder side flashblock delivery view and measurement of delivery latency between builder and flashblock websocket.

### DIFF
--- a/crates/op-rbuilder/src/builder/timing.rs
+++ b/crates/op-rbuilder/src/builder/timing.rs
@@ -1,5 +1,5 @@
 use core::time::Duration;
-use std::ops::Rem;
+use std::{ops::Rem, time::SystemTime};
 
 use reth_payload_builder::PayloadId;
 use tokio::sync::mpsc;
@@ -249,6 +249,17 @@ impl std::fmt::Debug for FlashblockScheduler {
             }))
             .finish()
     }
+}
+
+/// Compute the elapsed time in ms since the start of the slot.
+/// Slot start defined as `payload_timestamp - block_time`.
+pub(super) fn compute_slot_offset_ms(payload_timestamp: u64, block_time: Duration) -> f64 {
+    let slot_start = SystemTime::UNIX_EPOCH + Duration::from_secs(payload_timestamp) - block_time;
+    SystemTime::now()
+        .duration_since(slot_start)
+        .unwrap_or_default()
+        .as_secs_f64()
+        * 1000.0
 }
 
 #[cfg(test)]
--- a/crates/op-rbuilder/src/metrics.rs
+++ b/crates/op-rbuilder/src/metrics.rs
@@ -244,6 +244,22 @@ impl OpRBuilderMetrics {
     }
 }
 
+/// Record the slot-relative time at which a flashblock was published.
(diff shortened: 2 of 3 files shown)


crates/op-rbuilder/src/metrics.rs
```diff
@@ -232,6 +232,22 @@ impl OpRBuilderMetrics {
     }
 }
 
+/// Record the slot-relative time at which a flashblock was published.
+/// `offset_ms` is the time elapsed since slot start (payload_timestamp - block_time)
+pub fn record_flashblock_publish_timing(flashblock_index: u64, offset_ms: f64) {
+    static LABELS: &[&str] = &["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10"];
```
akundaz, line 238: instead of doing this you can just convert the index to a string

crates/op-rbuilder/src/metrics.rs
```diff
@@ -232,6 +232,22 @@ impl OpRBuilderMetrics {
...
+    let index_str = LABELS
+        .get(flashblock_index as usize)
+        .copied()
+        .unwrap_or("unknown");
+    let labels: [(&str, &str); 1] = [("flashblock_index", index_str)];
+    gauge!(
```
akundaz, line 244: we should stop using gauges for the most part since they get updated much more often than prometheus scrapes. a histogram would be more useful

crates/op-rbuilder/src/builder/timing.rs
```diff
@@ -250,6 +250,17 @@ impl std::fmt::Debug for FlashblockScheduler {
     }
 }
 
+/// Compute the elapsed time in ms since the start of the slot.
+/// Slot start defined as `payload_timestamp - block_time`.
+pub(super) fn compute_slot_offset_ms(payload_timestamp: u64, block_time: Duration) -> f64 {
```
akundaz, line 255: it would be cleaner to move this into the metrics function

avalonche (approved):

crates/op-rbuilder/src/builder/timing.rs
julio4: i find the current approach to be the cleanest one, timing module define what is slot start/offset of flashblock; metric module only handle recording metric values

crates/op-rbuilder/src/metrics.rs
julio4: the issue is that labels must be static str, so when converting the index on the heap you would need to continuously leak the same set of values.

avalonche: this is fragile no? when you have more than 10 flashblocks

julio4: More than 10 will just group themselves as 'unknown' label. ~Slightly fragile but not aware of a better solution. This is still better than leaking, even if very small (\~5MB per day)~
Actually we can use owned string within metrics macros as it handles cow str, I was using ref str hence need for static.

PR #466, opened 2026-04-17
feat: top-of-block pre-simulation to filter reverting tx spam

akundaz:
## 📝 Summary

Features:
- RPC layer filter: Bundle transactions marked as allowed to revert are now simulated against the current chain tip immediately following pool insertion. If they revert, then they're removed from the pool, saving compute from block building.
- Background pool eviction: A maintenance task listens for newly promoted pending transactions and evicts reverting ones, closing the nonce-gap bypass where adversaries could bypass RPC pre-simulation by entering their tx into the queued pool first.
- Random coinbase for simulation: Prevents attackers from crafting transactions that behave differently during simulation vs block building by checking block.coinbase.

Unit tests and a integration test added.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -61,6 +61,15 @@ pub struct OpRbuilderArgs {
     )]
     pub exclude_reverts_between_flashblocks: bool,
 
+    /// Whether to pre-simulate bundle transactions against the current head state
+    /// and reject them if they revert, before adding to the pool.
+    #[arg(
+        long = "builder.pre-simulate-bundles",
+        default_value = "false",
+        env = "PRE_SIMULATE_BUNDLES"
+    )]
+    pub pre_simulate_bundles: bool,
+
     /// Enables logs to trace transaction lifecycle as it is added to the
     /// mempool and added to a block. Requires `RUST_LOG=tx_trace=debug` in
     /// addition to this flag.
--- a/crates/op-rbuilder/src/presim.rs
+++ b/crates/op-rbuilder/src/presim.rs
@@ -1,0 +1,408 @@
+use std::sync::Arc;
+
+use alloy_consensus::{BlockHeader, Header};
+use alloy_evm::{Evm, InvalidTxError};
+use alloy_op_evm::OpTx;
+use alloy_primitives::{Address, B256, Bytes};
+use eyre::Context;
+use futures_util::{Stream, StreamExt};
+use parking_lot::RwLock;
+use reth_chain_state::CanonStateNotification;
+use reth_evm::{EvmError, IntoTxEnv};
+use reth_optimism_chainspec::{OpChainSpec, OpHardforks};
+use reth_optimism_evm::{OpEvmConfig, OpNextBlockEnvAttributes};
+use reth_optimism_primitives::OpTransactionSigned;
+use reth_primitives_traits::{Block, NodePrimitives, Recovered, RecoveredBlock};
+use reth_provider::{BlockReaderIdExt, ChainSpecProvider, StateProvider, StateProviderFactory};
+use reth_revm::{State, database::StateProviderDatabase};
+use reth_transaction_pool::{FullTransactionEvent, PoolTransaction, TransactionPool};
+use revm::{context::result::ResultAndState, context_interface::result::InvalidTransaction};
+use tracing::{debug, error, warn};
+
+use crate::{evm::OpBlockEvmFactory, metrics::OpRBuilderMetrics, tx::FBPooledTransaction};
+
+/// Pre-simulates transactions against the current head state to filter out
+/// reverting transactions before they enter the pool.
+pub(crate) struct TopOfBlockSimulator {
+    tip_state: RwLock<Arc<Option<TipState>>>,
+}
+
+pub(crate) struct TipState {
+    evm_factory: OpBlockEvmFactory,
+    state_provider: Box<dyn StateProvider + Send + Sync>,
+}
+
+impl TopOfBlockSimulator {
+    pub(crate) fn new() -> Self {
+        Self {
+            tip_state: RwLock::new(Arc::new(None)),
+        }
+    }
+
+    pub(crate) async fn simulate_tx(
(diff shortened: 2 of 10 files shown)


crates/op-rbuilder/src/args/op.rs
```diff
@@ -61,6 +61,15 @@ pub struct OpRbuilderArgs {
     )]
     pub exclude_reverts_between_flashblocks: bool,
 
+    /// Whether to pre-simulate bundle transactions against the current head state
+    /// and reject them if they revert, before adding to the pool.
+    #[arg(
+        long = "builder.pre-simulate-bundles",
```
avalonche, line 67: worth adding a simulation timeout / max gas usage config for bundles

crates/op-rbuilder/src/revert_protection.rs
```diff
@@ -141,7 +162,45 @@ where
...
             .map_err(EthApiError::from)?;
 
-        let recovered = recover_raw_transaction(&bundle_transaction)?;
+        let recovered: Recovered<op_alloy_consensus::OpPooledTransaction> =
+            recover_raw_transaction(&bundle_transaction)?;
+
+        // Pre-simulate the transaction against current head state if:
+        // - pre-simulation is enabled (simulator is Some)
+        // - the tx is allowed to revert
+        if bundle
```
avalonche, line 171: is min block / timestamp accounted for here?

crates/op-rbuilder/src/presim.rs
```diff
@@ -0,0 +1,409 @@
...
+            OpBlockEvmFactory::for_next_block(evm_config, tip.header(), &block_env_attributes)?;
+        let state_provider = provider.state_by_block_hash(tip.hash_slow())?;
+
+        // SAFETY: reth's concrete StateProvider implementations are Send + Sync.
+        // StateProviderBox omits these bounds but the underlying types satisfy them.
+        let state_provider: Box<dyn StateProvider + Send + Sync> =
+            unsafe { std::mem::transmute::<Box<dyn StateProvider>, _>(state_provider) };
+        // Compile-time assertion
+        const _: () = assert!(
+            std::mem::size_of::<Box<dyn StateProvider>>()
```
SozinM: But they also could have different layout, which would have the same size

crates/op-rbuilder/src/presim.rs
```diff
@@ -0,0 +1,409 @@
...
+            evm_factory,
+            state_provider,
+        })
+    }
+
+    fn run_simulation(&self, tx: impl IntoTxEnv<OpTx>) -> bool {
+        let db = StateProviderDatabase::new(&*self.state_provider);
+        let mut state = State::builder()
+            .with_database(db)
+            .with_bundle_update()
```
SozinM: You could skip bundle update if you only need failed/passed

crates/op-rbuilder/src/presim.rs
```diff
@@ -0,0 +1,409 @@
...
+
+        match evm.transact(tx) {
+            Ok(ResultAndState { result, .. }) => result.is_success(),
+            Err(err) => {
+                if err.as_invalid_tx_err().is_some_and(|err| {
+                    err.as_invalid_tx_err().is_some_and(|err| {
+                        matches!(
+                            err,
+                            InvalidTransaction::NonceTooLow { .. }
+                                | InvalidTransaction::NonceTooHigh { .. }
```
SozinM, line 131: I feel like this would be 90 majority of cases, and we could catch it early without tx execution, by checking nonces from db and exit function early if we see that we cannot execute it

crates/op-rbuilder/src/presim.rs
```diff
@@ -0,0 +1,409 @@
...
+                        )
+                    })
+                }) {
+                    // Nonce gap — tx may depend on pending state, let the pool handle it
```
SozinM, line 135: Iirc pool has a method to get ancestors of tx, might be useful

SozinM: Or you could take SimTree structure and keep if for simulations

crates/op-rbuilder/src/revert_protection.rs
akundaz: i moved the presim into a task that happens after we add it to the pool, so that the simulation doesn't block the rpc handler from responding. so the pool will check for things like block/timestamp, and the presim task will only run if it's successful now.

crates/op-rbuilder/src/presim.rs
akundaz: i added a check for alignment, do you think that should be enough?

crates/op-rbuilder/src/presim.rs
akundaz: done

crates/op-rbuilder/src/presim.rs
akundaz: that's probably true but wouldn't the evm check nonces before trying to execute anyway?

crates/op-rbuilder/src/presim.rs
SozinM: I think we just need to carefully check the docs about UB in transmute and ensure that it's okay (and reflect it in the doc) 
Because it might be completely safe to do as you did

crates/op-rbuilder/src/presim.rs
SozinM: You are right, my bad

crates/op-rbuilder/src/presim.rs
akundaz: i'm leaning towards it being safe, i'm not sure it's much different from this unsafe impl in signal boost https://github.com/flashbots/signal-boost/blob/a6ec461fee38e026d0f3e3429a09a6b94460eb9b/crates/signal-boost/src/cached_state.rs#L28

avalonche (approved):

crates/op-rbuilder/src/presim.rs
akundaz: I'm planning on doing the simtree in a follow up

PR #467, opened 2026-04-17
feat: default flashblocks.end-buffer-ms to 100

noot:
## 📝 Summary

The flashblock scheduler generates `target_flashblocks = block_time / interval` trigger times, with the last one firing at `slot_end - end_buffer_ms`. With the default `end_buffer_ms=0`, the final flashblock fires at the slot deadline itself and almost always fails to publish before `getPayload` — so the scheduler allocates a slot that is silently dropped, leaving N-1 usable flashblocks per block instead of N.

Benchmarked on a 2s chain block with the default 250ms interval (8 flashblocks/ block), 100 agents at 100M gas for 120s:

  end_buffer_ms  | p95 backrun_sent→landed | idx=8 utilized | landed/120s
  ---------------+-------------------------+----------------+-------------
       0         |        482 ms           |        no      |    34,686
      50         |        482 ms           |        no      |    35,976
     100         |        331 ms (-31%)    |       yes      |    40,936
     150         |        381 ms           |       yes      |    39,393

50ms is not enough buffer for the build/publish; 100ms is the smallest value in {50,100,150} that lets idx=8 publish before the slot deadline. Larger values (150ms+) don't help because the extra buffer becomes a larger end-of-block dead zone that forces more bundles into the next block.

100ms of buffer costs ~5% of the final slot's bundle-intake window (100ms of 2000ms) in exchange for the entire slot being usable. Net effect: 31% lower p95 landing latency and 18% higher per-block throughput.

## 💡 Motivation and Context

Improves signal-boost backrun tail latencies.

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -194,13 +194,20 @@ pub struct FlashblocksArgs {
     )]
     pub flashblocks_send_offset_ms: i64,
 
-    /// Time in milliseconds to build the last flashblock early before the end of the slot
+    /// Time in milliseconds to build the last flashblock early before the end of the slot.
     /// This serves as a buffer time to account for the last flashblock being delayed
-    /// at the end of the slot due to processing the final block
+    /// at the end of the slot due to processing the final block.
+    ///
+    /// Default 100ms: with the 250ms flashblock interval default and a 2s block, the last
+    /// flashblock is scheduled at t=slot_end. With zero buffer it typically can't publish
+    /// before `getPayload` and is silently dropped, leaving one of the block's flashblock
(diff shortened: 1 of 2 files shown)


akundaz:
This doesn't need to be a default though right? We don't use 0 either for running the builder but we don't want to expose what value we're using since adversaries would then know by how much they'd need to degrade building. Also, with continuous building (https://github.com/flashbots/op-rbuilder/pull/458) I think `getPayload` will trigger the last flashblock send (as it's already built) so then a 0 default is fine

julio4 (approved):

noot:
@akundaz yeah, if the value used in prod isn't 0 then that should be fine, i'll re-bench with continuous building also then

PR #472, opened 2026-04-22
fix(presim): only simulate bundles with revert protection

akundaz:


crates/op-rbuilder/src/presim.rs
```diff
@@ -196,6 +196,11 @@ pub(crate) async fn maintain_pending_simulations<Pool, St>(
             continue;
         };
 
+        // Only simulate txs that enable revert protection
+        if tx.transaction.reverted_hashes().is_none() {
```
avalonche, line 200: if the bundle specifies its hashes to be reverting, this will reject bundles where the tx hash is allowed to revert and land on chain?

avalonche (commented):
please add more unit tests for this behaviour

crates/op-rbuilder/src/presim.rs
akundaz: I don't understand this question. We only want to simulate and potentially reject bundles that enable revert protection, since otherwise the sender would be paying gas fees.

akundaz:
> please add more unit tests for this behaviour

I'm struggling to find a way to unit test this that would add value. What behavior did you want to see checked?

crates/op-rbuilder/src/presim.rs
avalonche: I mean the revertingTxHashes explicitly allows these transaction hashes to revert  https://developers.uniswap.org/docs/unichain/technical-information/advanced-txn#method-signature so technically they don't have revert protection and don't need to go through the presim path

avalonche:
check that reverted_hashes is None and reverted_hashes is Some is correctly filtered? Also tx hashes in revertingTxHashes is not discarded
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_sendBundle",
  "params": [
    {
      "txs": ["0x..."],                    // Array with single tx
      "minBlockNumber": "0x...",           // Optional: minimum block number (hex)
      "maxBlockNumber": "0x...",           // Optional: maximum block number (hex)
      "minTimestamp": 1234567890,          // Optional: minimum timestamp (number)
      "maxTimestamp": 1234567899,          // Optional: maximum timestamp (number)
      "revertingTxHashes": ["0x..."]       // Optional: transactions allowed to revert
    }
  ]
}
```

crates/op-rbuilder/src/presim.rs
akundaz: are you saying the `if tx.transaction.reverted_hashes().is_none()` check is incorrect? i'm trying to skip the presim path for for txs that don't opt in to revert protection

avalonche: you just need to check that its empty instead of none I think

akundaz: updated this so it's consistent with checking revert protection during building

avalonche (approved):

akundaz:
> check that reverted_hashes is None and reverted_hashes is Some is correctly filtered? Also tx hashes in revertingTxHashes is not discarded

I'll work this out in a follow up

PR #473, opened 2026-04-22
hotfix: point to op-reth branch to use v3 payload id instead of v4

avalonche:
## 📝 Summary

Point to hot fix branch of op-reth to use v3 payload instead of v4.

## 💡 Motivation and Context

payload id mismatch between op-geth and op-reth. Temporary patch, but long term rollup-boost should handle the payload id mismatch between builder and local l2 builder. 

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,55 +56,55 @@ unused_async = "warn"
 
 [workspace.dependencies]
 # reth
-reth = { git = "https://github.com/paradigmxyz/reth", rev = "082c36e" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "0.4.2"
+version = "0.4.3-payload-id-fix.1"
```
akundaz: the version should be managed by the release script not updated manually

avalonche: this is a hot fix tag

julio4 (approved):

PR #480, opened 2026-04-27
refactor(generator): behaviour fixes for new_payload_job and best_payload

julio4:
## Summary

Stacked on top of #479

### FCU arrival-delay metric
- Use `payload_time.checked_sub(self.block_time).unwrap_or(UNIX_EPOCH)` against underflow when `block_time` exceeds payload timestamp.
- Replace `Duration → u128 → i64 → f64` cast chain with single `Duration::as_secs_f64() * 1000.0`.

### `new_payload_job`
- Take only once lock on `last_payload_cancel`. Fire the cancel on the previous job and store the new cancellation token in single critical section.
- parent-header lookup moved **before** cancellation step: a malformed FCU (bad `parent_hash`) no longer aborts in-progress work.

### `resolve_kind`
- drop `payload_kind` log field. The kind is never used.

### `best_payload`
- Provide a minimal implementation instead of `unimplemented`.

### DIFF
--- a/crates/op-rbuilder/src/builder/generator.rs
+++ b/crates/op-rbuilder/src/builder/generator.rs
@@ -149,32 +149,15 @@ where
         // Expected: FCU should arrive at (payload_timestamp - block_time)
         // Positive delay = FCU arrived late, Negative = FCU arrived early
         let timestamp = rpc_attributes.timestamp();
-        let now = SystemTime::now();
-        let expected_fcu_arrival =
-            SystemTime::UNIX_EPOCH + Duration::from_secs(timestamp) - self.block_time;
-        let fcu_arrival_delay_ms = now
-            .duration_since(expected_fcu_arrival)
-            .map(|d| d.as_millis() as i64)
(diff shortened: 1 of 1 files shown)


crates/op-rbuilder/src/builder/generator.rs
```diff
@@ -287,9 +284,9 @@ where
 
     fn resolve_kind(
         &mut self,
-        kind: PayloadKind,
+        _kind: PayloadKind,
```
akundaz, line 287: why remove this from the log? might as well keep it

akundaz (approved):

crates/op-rbuilder/src/builder/generator.rs
julio4: We don't use it at all, so logging the kind is pretty useless. PayloadKind is only a hint to the builder to either send payload as early as possible or ok with waiting a bit more (1s deadline), it is not very relevant here.

PR #481, opened 2026-04-27
feat(presim): constrain number of sim tasks

akundaz:
1st commit: refactoring pool insertion to happen after sending on a channel. This will make it easier later on to put the presim at the right spot (before bundles get into the reth pool).

2nd commit: Added a flag to constrain the number of simultaneously running presim tasks

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -70,6 +70,14 @@ pub struct OpRbuilderArgs {
     )]
     pub pre_simulate_bundles: bool,
 
+    /// Maximum number of concurrent pre-simulation tasks. If not set,
+    /// concurrency is unbounded.
+    #[arg(
+        long = "builder.presim-max-concurrency",
+        env = "PRESIM_MAX_CONCURRENCY"
+    )]
+    pub presim_max_concurrency: Option<usize>,
+
     /// Enables logs to trace transaction lifecycle as it is added to the
     /// mempool and added to a block. Requires `RUST_LOG=tx_trace=debug` in
     /// addition to this flag.
--- a/crates/op-rbuilder/src/pool.rs
+++ b/crates/op-rbuilder/src/pool.rs
@@ -1,0 +1,73 @@
+use std::{sync::Arc, time::Instant};
+
+use reth_tasks::Runtime;
+use reth_transaction_pool::{
+    PoolTransaction, TransactionOrigin, TransactionPool, error::PoolError,
+};
+use tokio::sync::{mpsc, oneshot};
+use tracing::error;
+
+use crate::{metrics::OpRBuilderMetrics, presim::TopOfBlockSimulator, tx::FBPooledTransaction};
+
+#[derive(derive_more::Constructor)]
+pub(super) struct AddBundleRequest {
+    tx: FBPooledTransaction,
+    respond_to: oneshot::Sender<Option<PoolError>>,
+}
+
+pub(super) async fn run_pool_insertion_task(
+    runtime: Runtime,
+    // TODO: Remove 'static bound
+    pool: impl TransactionPool<Transaction = FBPooledTransaction> + 'static,
+    mut rx: mpsc::UnboundedReceiver<AddBundleRequest>,
+    simulator: Option<Arc<TopOfBlockSimulator>>,
+    metrics: Arc<OpRBuilderMetrics>,
+) {
+    while let Some(req) = rx.recv().await {
+        let AddBundleRequest { tx, respond_to } = req;
+
+        if let Err(e) = pool
+            .add_transaction(TransactionOrigin::Local, tx.clone())
+            .await
+        {
+            let _ = respond_to.send(Some(e));
+            continue;
+        } else {
+            let _ = respond_to.send(None);
+        }
+
+        // Pre-simulate the transaction against current head state if:
+        // - pre-simulation is enabled (simulator is Some)
+        // - the tx has opted into revert protection
+        if tx.revert_protected()
+            && let Some(simulator) = &simulator
+        {
+            let pool = pool.clone();
+            let metrics = metrics.clone();
+            let simulator = simulator.clone();
+            runtime.spawn_task({
+                let runtime = runtime.clone();
+                async move {
+                    let sim_start = Instant::now();
+                    let sim_tx = tx.into_consensus();
+                    let sim_tx_hash = *sim_tx.hash();
+                    match simulator.clone().simulate_tx(&runtime, sim_tx).await {
+                        Ok(true) => {
+                            metrics.bundle_pre_simulation_passes.increment(1);
+                        }
+                        Ok(false) => {
+                            metrics.bundle_pre_simulation_reverts.increment(1);
+                            pool.remove_transaction(sim_tx_hash);
+                        }
+                        Err(e) => {
+                            error!(error = %e, "pre-simulation task failed");
+                        }
+                    }
+                    metrics
+                        .bundle_pre_simulation_duration
+                        .record(sim_start.elapsed());
+                }
+            });
+        }
+    }
+}
--- a/crates/op-rbuilder/src/pool.rs
+++ b/crates/op-rbuilder/src/pool.rs
@@ -1,10 +1,12 @@
 use std::{sync::Arc, time::Instant};
 
+use futures::StreamExt;
 use reth_tasks::Runtime;
 use reth_transaction_pool::{
     PoolTransaction, TransactionOrigin, TransactionPool, error::PoolError,
 };
 use tokio::sync::{mpsc, oneshot};
+use tokio_stream::wrappers::ReceiverStream;
 use tracing::error;
 
 use crate::{metrics::OpRBuilderMetrics, presim::TopOfBlockSimulator, tx::FBPooledTransaction};
@@ -22,7 +24,19 @@ pub(super) async fn run_pool_insertion_task(
     mut rx: mpsc::UnboundedReceiver<AddBundleRequest>,
     simulator: Option<Arc<TopOfBlockSimulator>>,
     metrics: Arc<OpRBuilderMetrics>,
+    max_concurrency: Option<usize>,
 ) {
+    let bounded_sender = max_concurrency.map(|max| {
+        let (sender, receiver) = mpsc::channel(max);
+        runtime.spawn_task(async move {
+            let _ = ReceiverStream::new(receiver)
+                .buffer_unordered(max)
+                .for_each(|_| async {})
+                .await;
+        });
+        sender
+    });
+
     while let Some(req) = rx.recv().await {
         let AddBundleRequest { tx, respond_to } = req;
 
@@ -45,7 +59,7 @@ pub(super) async fn run_pool_insertion_task(
             let pool = pool.clone();
             let metrics = metrics.clone();
             let simulator = simulator.clone();
-            runtime.spawn_task({
+            let fut = {
                 let runtime = runtime.clone();
                 async move {
                     let sim_start = Instant::now();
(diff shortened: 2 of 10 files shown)


crates/op-rbuilder/src/pool.rs
```diff
@@ -0,0 +1,55 @@
...
+use reth_transaction_pool::{PoolTransaction, TransactionOrigin, TransactionPool};
+use tokio::sync::mpsc;
+use tracing::error;
+
+use crate::{metrics::OpRBuilderMetrics, presim::TopOfBlockSimulator, tx::FBPooledTransaction};
+
+pub(super) async fn run_pool_insertion_task(
+    // TODO: Remove 'static bound
+    pool: impl TransactionPool<Transaction = FBPooledTransaction> + 'static,
+    mut rx: mpsc::UnboundedReceiver<FBPooledTransaction>,
```
julio4: We should add a bound to this pre-sim pool as send_bundle rpc endpoint may be spammed. We can start high and monitor performance.

crates/op-rbuilder/src/revert_protection.rs
```diff
@@ -167,55 +144,17 @@ where
...
-                    }
-                }
-                metrics
-                    .bundle_pre_simulation_duration
-                    .record(sim_start.elapsed());
-            });
-        }
+        self.pool_tx
+            .send(pool_transaction)
+            .map_err(|_| EthApiError::InternalEthError)?;
```
julio4, line 158: The transaction is a raw unvalidated rpc call and should return RpcResult<BundleResult>::EthApiError if validation fails, but here these errors are not propagated back here after being sent to the pool insertion task (only err handled is channel communication). Consider add a oneshot response and await for rpc response, but this will add latency to the response and I'm not sure what is acceptable.

crates/op-rbuilder/src/pool.rs
```diff
@@ -0,0 +1,55 @@
...
+    simulator: Option<Arc<TopOfBlockSimulator>>,
+    metrics: Arc<OpRBuilderMetrics>,
+) {
+    while let Some(tx) = rx.recv().await {
+        if let Err(e) = pool
+            .add_transaction(TransactionOrigin::Local, tx.clone())
+            .await
+        {
+            error!(error = ?e, "failed to add bundle to pool");
+        }
```
julio4, line 51: If validation fails, it should stop loop iteration here and not continue to presim.

crates/op-rbuilder/src/pool.rs
```diff
@@ -0,0 +1,55 @@
...
+        // Pre-simulate the transaction against current head state if:
+        // - pre-simulation is enabled (simulator is Some)
+        // - the tx has opted into revert protection
+        if tx.revert_protected()
+            && let Some(simulator) = &simulator
+        {
+            let pool = pool.clone();
+            let metrics = metrics.clone();
+            let simulator = simulator.clone();
+            tokio::task::spawn(async move {
```
julio4: Do not directly spawn tokio task outside of reth task manager/`Runtime`.
Also, I'm not sure if another task make sense here as we may need the result to officially respond to initial rpc call.

crates/op-rbuilder/src/pool.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::{metrics::OpRBuilderMetrics, presim::TopOfBlockSimulator, tx::FBPooledTransaction};
+
+pub(super) async fn run_pool_insertion_task(
+    // TODO: Remove 'static bound
+    pool: impl TransactionPool<Transaction = FBPooledTransaction> + 'static,
+    mut rx: mpsc::UnboundedReceiver<FBPooledTransaction>,
+    simulator: Option<Arc<TopOfBlockSimulator>>,
+    metrics: Arc<OpRBuilderMetrics>,
+) {
+    while let Some(tx) = rx.recv().await {
```
julio4: I don't think we have fifo processing guarantees, so we could also think about other patterns such as small pool of workers instead of one sequential loop.

crates/op-rbuilder/src/revert_protection.rs
akundaz: no this will add way too much latency to the response

crates/op-rbuilder/src/pool.rs
akundaz: no? it doesn't indicate something broken with presim

crates/op-rbuilder/src/pool.rs
akundaz: spawning a new task makes the most sense since i actually don't want the result included in the rpc call

crates/op-rbuilder/src/pool.rs
akundaz: the sequential loop is fine because the work done in the loop body is very minimal and shouldn't apply backpressure to the rpc handler. this isn't about fifo guarantees it's just the simplest implementation

crates/op-rbuilder/src/pool.rs
akundaz: updated to use the reth runtime

crates/op-rbuilder/src/pool.rs
akundaz: i'm not sure what you mean by bound in this case but eventually yes i do think we should constrain how many simulations we're running at once

crates/op-rbuilder/src/revert_protection.rs
julio4: We still need to return the first validation part. Also when sim fails and bundle is not added to the pool, the rpc response has success status and there's no way to know that it was not included in the pool.

crates/op-rbuilder/src/pool.rs
julio4: why would you simulate a invalid bundle?

akundaz: oh i see. just added a `continue` statement after logging the error

crates/op-rbuilder/src/revert_protection.rs
akundaz: my plan is to later update the reverted cache so the user can call `eth_getTransactionReceipt` to check if their tx reverted.

i'll add a oneshot but i'm going to have to get rid of it really soon because the presim should really run before we add a tx to the pool

crates/op-rbuilder/src/pool.rs
julio4: Validation is cheap but not simulation, so we should add some back-pressure else we can spawn an unbounded amount of sim tasks and starve builder no?
And yes I agree one sequential loop is enough for now for simplicity, just thinking out loud if it becomes a bottleneck during high bundles submission time.

akundaz: right now the current implementation does not simulate every incoming bundle. once i get the correct implementation working with a simtree then it would make sense to optimize for performance.

crates/op-rbuilder/src/pool.rs
```diff
@@ -0,0 +1,61 @@
...
+        // Pre-simulate the transaction against current head state if:
+        // - pre-simulation is enabled (simulator is Some)
+        // - the tx has opted into revert protection
+        if tx.revert_protected()
+            && let Some(simulator) = &simulator
+        {
+            let pool = pool.clone();
+            let metrics = metrics.clone();
+            let simulator = simulator.clone();
+            runtime.spawn_task({
```
SozinM: this will blow up

crates/op-rbuilder/src/pool.rs
julio4: With this then yes agree!

crates/op-rbuilder/src/pool.rs
SozinM: better to spawn N persistent tasks, that will receive txs via channel, and then insert into the pool if tx is good

crates/op-rbuilder/src/pool.rs
SozinM: in there i'd suggest use recv_many and use reth pool ability to add multiple txs

crates/op-rbuilder/src/pool.rs
akundaz: i'm aware of this. but this is a refactoring pr and i think implementing a worker pool is out of scope at the moment

crates/op-rbuilder/src/revert_protection.rs
julio4: Yeah I see, it's a bit tricky as the validation takes part within the add_to_pool and then simulation removes from the pool when failing. The best would be to decouple validation from adding to pool and on sendBundle call directly return invalid/pending(simulation).

crates/op-rbuilder/src/pool.rs
julio4: Use a bounded channel with a oneshot response for validation. But without a worker pool it'll be harder because you don't await the sim task.

akundaz: what does this achieve? the simulation doesn't apply any backpressure so a bounded channel won't constrain our compute resources

crates/op-rbuilder/src/pool.rs
akundaz: can you elaborate? is this a blocking concern for this pr?

SozinM: Nice cheap improvement, could add to this pr

akundaz: done. added a commit for this

crates/op-rbuilder/src/pool.rs
akundaz: this is done now

crates/op-rbuilder/src/args/op.rs
```diff
@@ -70,6 +70,14 @@ pub struct OpRbuilderArgs {
     )]
     pub pre_simulate_bundles: bool,
 
+    /// Maximum number of concurrent pre-simulation tasks. If not set,
+    /// concurrency is unbounded.
+    #[arg(
+        long = "builder.presim-max-concurrency",
+        env = "PRESIM_MAX_CONCURRENCY"
+    )]
+    pub presim_max_concurrency: Option<usize>,
```
julio4, line 79: Need to validate >0 as channel(0) would panic.

crates/op-rbuilder/src/pool.rs
```diff
@@ -0,0 +1,94 @@
...
+                        }
+                    }
+                    metrics
+                        .bundle_pre_simulation_duration
+                        .record(sim_start.elapsed());
+                }
+            };
+
+            match &bounded_sender {
+                Some(sender) => sender.send(fut).await.unwrap(),
```
julio4, line 87: When presim queue is full the send(fut).await blocks the single insertion worker so new RPC requests would pile up in the unbounded pool_rx. Consider using try_send and handle the TrySendError::Full with atleast a metric or log.

crates/op-rbuilder/src/pool.rs
```diff
@@ -0,0 +1,94 @@
...
+    pool: impl TransactionPool<Transaction = FBPooledTransaction> + 'static,
+    mut rx: mpsc::UnboundedReceiver<AddBundleRequest>,
+    simulator: Option<Arc<TopOfBlockSimulator>>,
+    metrics: Arc<OpRBuilderMetrics>,
+    max_concurrency: Option<usize>,
+) {
+    let bounded_sender = max_concurrency.map(|max| {
+        let (sender, receiver) = mpsc::channel(max);
+        runtime.spawn_task(async move {
+            let _ = ReceiverStream::new(receiver)
```
julio4, line 32: Maybe a PresimJob struct would be clearer instead of sending future through the channel, and you create the future here.

julio4 (commented):
I have a slight feeling that this whole module adds some unnecessary complexity. In the current form it would be better to keep validation + response at the rpc level and only spawn this "pool presim" worker as a post-hook after pool insertion. If the goal is to have an invariant every bundle need to be simulated before being added to the pool, then it would be better to first implement changes to make that possible.

akundaz:
Closing this PR as there's too much feature creep going on

PR #486, opened 2026-05-02
fix: propagate eth_api errors instead of panicking

kevzzsk:
## 📝 Summary

`transaction_receipt` unwrapped the upstream `eth_api.transaction_receipt` result, so any error from the underlying call (DB hiccup, provider blip, etc.) panicked the RPC handler instead of being returned to the caller.

---

* [x] Run `make lint`
* [x] Run `make test`


julio4 (approved):
Thanks LGTM! Simplified it a bit as I don't see the need to extract logic in a separate helper yet.

PR #487, opened 2026-05-04
feat: per-sender simulation time limiting

akundaz:
### Per-sender compute time limiter
This new rate limiter allocates an amount of simulation time for each sender's order flow. This should help us throttle more accurately, especially if adversaries are using underpriced EVM precompiles. 

Operation is very similar to the gas limitier -- it's configurable via CLI args `--compute-limiter.*` vs `--gas-limiter.*`:
- `--compute-limiter.enabled`
- `--compute-limiter.max-time-us-per-address`
- `--compute-limiter.refill-rate-us-per-block`
- `--compute-limiter.cleanup-interval`

### Move rate limiting checks before EVM execution
Previously we would check if a bucket had enough tokens in it before deciding whether to commit the tx to the block or not. This is a bit of a bug, since it requires simulating the tx, which we'd like to avoid.

So this PR introduces a new concept of **debt**, tracked in each bucket. Each time we simulate a tx, we charge the gas used and compute time to the sender's buckets, which can now go "negative", or in debt. If a sender is in debt, we skip simulating their order flow entirely.

Caveat: this means we allow a free overshoot per each sender. I don't think this is too big a deal, and we can add a check post-simulation to skip committing to the block if we really want to later. 

### etc.
- Refactors the gas_limiter module into limiter, extracting generic TokenBucket<T> and AddressBuckets<T> types that are reused by both the gas limiter (u64) and compute limiter (Duration).
- Deprecates the timing metrics since we never use them and they're too insignificant

There's been discussion about only charging for reverted txs -- with this PR it's a trivial change of moving the `consume_gas` and `consume_compute` calls.


julio4 (commented):
This is an interesting idea, but I have a few concerns with using per-sender simulation time as a budget:
- stateful per-sender tracking adds overhead. When running concurrent candidate builds in isolation, copying/cloning these mappings can slowly adds up.
- “Compute time” is builder-dependent and opaque to users. Unlike gas it’s not a transparent or protocol-level metric.
- It’s also not fully deterministic with respect to the input even if variance should be small.

An simpler alternative could be stateless per-transaction timeout, and potentially apply rate limiting to senders that repeatedly exceed it.

akundaz:
> stateful per-sender tracking adds overhead. When running concurrent candidate builds in isolation, copying/cloning these mappings can slowly adds up.

Do you have data on this? How much does it add up to?

Any kind of rate limiting on senders will have to be stateful. This isn't a departure from the existing gas limiter.

> “Compute time” is builder-dependent and opaque to users. Unlike gas it’s not a transparent or protocol-level metric.

I don't see the problem with this

> It’s also not fully deterministic with respect to the input even if variance should be small.

Again not a problem.  And this PR includes the machinery to easily add rate limiting based on other parameters later on if we want.

> An simpler alternative could be stateless per-transaction timeout, and potentially apply rate limiting to senders that repeatedly exceed it.

This is more complex solution than what's in this PR

avalonche:
alternatively we can use CPU time to be more accurate when builder is under load, it'll be good to measure both

crates/op-rbuilder/src/limiter/metrics.rs
```diff
@@ -0,0 +1,18 @@
+use metrics::{Counter, Gauge};
+use reth_metrics::Metrics;
+
+#[derive(Metrics, Clone)]
+#[metrics(scope = "op_rbuilder")]
+pub(super) struct LimiterMetrics {
```
avalonche, line 6: histogram of the amount of "debt" would be useful

crates/op-rbuilder/src/primitives/reth/execution.rs
```diff
@@ -19,6 +19,7 @@ pub enum TxnExecutionResult {
     Success,
     Reverted,
     RevertedAndExcluded,
+    SenderBudgetExhausted,
     MaxGasUsageExceeded,
```
avalonche, line 23: `MaxGasUsageExceeded` replaced by `SenderBudgetExhausted` can be removed

avalonche (approved):

avalonche:
we can add a timeout in the revm Inspector to resolve the issue of an attacker constructing new addresses and sending 1 transaction with long simulation time, though that will need benchmarking

PR #488, opened 2026-05-05
refactor(gas_limiter): canonical + overlay design with epoch tag

julio4:
Refactor the per-address gas/compute limiter to use per-payload overlays over a shared canonical limiter.

  Each payload job now records limiter usage in its own overlay and commits it back when the job finishes. This keeps in-flight payload builds isolated while preserving persistent sender debt across jobs. The
  limiter also rejects stale overlay commits when the canonical state has already advanced.

  ## Changes

  - Add canonical + overlay limiter model for gas and compute buckets.
  - Add epoch checks to prevent stale overlays from committing.
  - Add checkpoint/restore support for rolling back failed tx attempts.
  - Wire payload jobs to use limiter overlays.
  - Refresh limiter state before forking payload overlays.


akundaz:
This conflicts hugely with my existing PR (https://github.com/flashbots/op-rbuilder/pull/487). Can you describe to me
* the api you need for continuous building
* and the performance concerns you have

Then we can build on top of each other and streamline the work properly

julio4:
> This conflicts hugely with my existing PR (#487). Can you describe to me
> 
> * the api you need for continuous building
> * and the performance concerns you have
> 
> Then we can build on top of each other and streamline the work properly

Yes that's why I extracted it from continuous PR, so that we can merge it first and we can update 487 with this foundation.

There's detailed "What was missing" and "What this fixes for continuous mode" sections in the description upper, but tldr we need cheap copy/overlay as only the winning candidate should consume budgets.

crates/op-rbuilder/src/limiter/bucket.rs
```diff
@@ -16,48 +16,156 @@ impl<T: Copy + Ord + Default + Add<Output = T> + AddAssign + Sub<Output = Self>
...
+            base: Arc::new(HashMap::new()),
+        }
+    }
+
     pub(super) fn len(&self) -> usize {
-        self.0.len()
+        self.base.len()
     }
 
+    /// Fork an overlay sharing this canonical's base. O(1).
```
akundaz: Rename methods for clarity:
- `refresh` -> `refill_buckets` so it can't be confused with the overlay concept
- `fork` -> `begin`
- `fold_overlay` -> `commit`
Makes it more clear this is basically a db transaction

crates/op-rbuilder/src/limiter/mod.rs
```diff
@@ -56,29 +176,63 @@ impl AddressLimiter {
...
+            l.consume_compute(address, time_used);
         }
     }
 
-    /// Should be called upon each new block. Refills buckets/Garbage collection
-    pub fn refresh(&self, block_number: u64) {
-        if let Some(inner) = self.gas_limiter.as_ref() {
-            inner.refresh(block_number)
+    /// Snapshot the current overlay state for later [`Self::restore`].
+    pub fn checkpoint(&self) -> AddressLimiterCheckpoint {
```
akundaz: I don't think you need to include `checkpoint` and `restore` methods in the public api. If you want to introduce a snapshot system then the `AddressBuckets` type at the bottom should be able to manage its own snapshots

crates/op-rbuilder/src/limiter/gas.rs
```diff
@@ -1,58 +1,90 @@
...
+        self.metrics
+            .gas_limiter_active_address_count
+            .set(self.gas_buckets.len() as f64);
+    }
+}
+
+/// In-flight overlay. Reads/consumes go through here with `&self`; on
+/// commit the accumulated deltas are folded back into the canonical.
+#[derive(Debug)]
+pub(super) struct GasLimiterOverlay {
```
akundaz: Can you work on following the generic pattern from bucket.rs when possible? It looks like there don't need to be separate `GasLimiterOverlay` and `ComputeLimiterOverlay` definitions

crates/op-rbuilder/src/builder/context.rs
```diff
@@ -93,6 +94,12 @@ pub struct OpPayloadJobCtx {
     pub cancel: CancellationToken,
     /// Per-block view into the global backrun bundle pool.
     pub backrun_pool: Option<BackrunBundlePayloadPool>,
+    /// Per-build overlay onto the canonical [`AddressLimiter`] held by
+    /// `builder_ctx`. Charges accumulate privately here and auto-commit
+    /// back into the canonical when this ctx is dropped. Shadows the
+    /// `address_limiter` field reached via `Deref` to `OpPayloadBuilderCtx`,
+    /// so `self.address_limiter` inside this ctx always hits the overlay.
+    pub address_limiter: AddressLimiterOverlay,
```
akundaz: An overlay is something that sits on top of something else to provide more information or context, but this is really a set of changes that are supposed to be atomically applied (or discarded) like a database transaction. I think naming it that way would help a lot with comprehension but since transaction is already taken can you rename the suffix to maybe `Guard`, `Scope`, `Lease`, or something else?

crates/op-rbuilder/src/limiter/bucket.rs
```diff
@@ -67,14 +175,13 @@ impl<T> Default for AddressBuckets<T> {
 /// Buckets can go into debt — `consume` always succeeds, but the bucket
 /// tracks how much was overdrawn.
 #[derive(Debug, Clone)]
-struct TokenBucket<T> {
+pub(super) struct TokenBucket<T> {
```
akundaz: Other modules shouldn't need to carry around and know about `TokenBucket`s, just the higher up `AddressBuckets` and the session type, please restrict this

akundaz (changes requested):

crates/op-rbuilder/src/builder/context.rs
julio4: I decided on Guard as there's some behavior on drop

akundaz (commented):
Looks much better thank you! Just get CI passing and I think this should be good to go

akundaz (approved):

PR #492, opened 2026-05-05
feat(metrics): Add labels to tx simulation metrics

avalonche:
##  Summary                                                   
                                                                                                                                         
  Replaces the unlabeled op_rbuilder_tx_simulation_duration histogram with a labeled version under the same metric name, so we can break 
  down per-tx EVM simulation latency along the dimensions that matter for tuning the builder hot path:
                                                                                                                                         
  - source — bundle | mempool                               
  - result — success | revert
  - revert_protected — true | false (always false for mempool txs; for bundles, true when the tx hash isn't in the bundle's              
  allowed-revert list)                                                                                                                   
                                                                                                                                         
 ## Compatibility                                                                                                                          
  
  The metric name is unchanged (op_rbuilder_tx_simulation_duration), but it now carries three labels.

### DIFF
--- a/crates/op-rbuilder/src/builder/context.rs
+++ b/crates/op-rbuilder/src/builder/context.rs
@@ -37,7 +37,9 @@ use crate::{
     evm::OpBlockEvmFactory,
     gas_limiter::AddressGasLimiter,
     hardforks::ActiveHardforks,
-    metrics::{OpRBuilderMetrics, record_tx_simulation_duration},
+    metrics::{
+        OpRBuilderMetrics, TxResult, TxSource, record_tx_gas_used, record_tx_simulation_duration,
+    },
     primitives::reth::{ExecutionInfo, TxnExecutionResult},
     traits::PayloadTxsBounds,
 };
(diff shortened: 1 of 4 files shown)


akundaz:
Can you change source to differentiate backruns from bundles?

Also, can you do the same thing for gas used?

You can deprecate some metrics if you do this:
* `successful_tx_gas_used`
* `reverted_tx_gas_used`
and maybe others that I'm not thinking of

crates/op-rbuilder/src/builder/context.rs
```diff
@@ -561,18 +579,36 @@ impl OpPayloadJobCtx {
...
                 num_txs_simulated_fail += 1;
                 reverted_gas_used += gas_used;
-                self.metrics.reverted_tx_gas_used.record(gas_used as f64);
-                if is_bundle_tx {
-                    num_bundles_reverted += 1;
+                match tx_source {
+                    TxSource::Mempool => match tx_result {
+                        TxResult::Revert => num_mempool_txs_simulated_revert += 1,
+                        TxResult::Halt => num_mempool_txs_simulated_halt += 1,
+                        TxResult::Success => {
```
akundaz, line 598: i think you can avoid the `unreachable!` invocations if you matched on `TxResult` instead of branching on `result.is_success()` above

crates/op-rbuilder/src/builder/context.rs
```diff
@@ -387,10 +389,17 @@ impl OpPayloadJobCtx {
     ) -> Result<Option<()>, PayloadBuilderError> {
         let execute_txs_start_time = Instant::now();
         let mut num_txs_considered = 0;
-        let mut num_txs_simulated = 0;
         let mut num_txs_simulated_success = 0;
         let mut num_txs_simulated_fail = 0;
-        let mut num_bundles_reverted = 0;
+        let mut num_mempool_txs_simulated_success = 0;
```
akundaz, line 394: just store all these in a hashmap or something so you just have one mutable variable for recording metrics. you wouldn't have so much repetitive code in `set_payload_builder_metrics` then either

PR #495, opened 2026-05-06
feat: create reth pool wrapper

akundaz:
## Summary
Create a new `Flashpool` to wrap reth's transaction pool.

I do this by creating a `FlashpoolBuilder` to wrap `OpPoolBuilder` and implement `PoolBuilder` by delegating to the inner `OpPoolBuilder`. The `Flashpool` then wraps around the `OpTransactionPool` created by the inner builder.

Implementing `TransactionPool` for `Flashpool` is done through the `delegate` macro for the 50+ methods on that trait. This will allow us to selectively customize individual methods later and access newly added fields on `Flashpool`.

## Motivation
The following features benefit from tighter integration with the pool:
* presim
* backruns
* multi-tx bundles
* reverted tx tracking
* .. and probably more that I can't think of right now

So while this is over-engineering at the present moment, I think this is the least unreasonable way to extend pool functionality.

### DIFF
--- a/crates/op-rbuilder/src/pool/builder.rs
+++ b/crates/op-rbuilder/src/pool/builder.rs
@@ -1,0 +1,47 @@
+use reth_evm::ConfigureEvm;
+use reth_node_api::{FullNodeTypes, NodeTypes, PrimitivesTy, TxTy};
+use reth_node_builder::{BuilderContext, components::PoolBuilder};
+use reth_optimism_forks::OpHardforks;
+use reth_optimism_node::OpPoolBuilder;
+use reth_optimism_txpool::{OpPooledTx, OpTransactionPool};
+use reth_transaction_pool::{EthPoolTransaction, blobstore::DiskFileBlobStore};
+
+use crate::{args::OpRbuilderArgs, pool::Flashpool, tx::FBPooledTransaction};
+
+pub struct FlashpoolBuilder {
+    op_pool_builder: OpPoolBuilder<FBPooledTransaction>,
+}
+
+impl FlashpoolBuilder {
+    pub fn new(builder_args: &OpRbuilderArgs) -> Self {
+        let rollup_args = &builder_args.rollup_args;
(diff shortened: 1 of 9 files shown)


crates/op-rbuilder/src/pool/builder.rs
```diff
@@ -0,0 +1,51 @@
...
+impl FlashpoolBuilder {
+    pub fn new(builder_args: &OpRbuilderArgs) -> Self {
+        let rollup_args = &builder_args.rollup_args;
+        let op_pool_builder = OpPoolBuilder::<FBPooledTransaction>::default()
+            .with_enable_tx_conditional(
+                // Revert protection uses the same internal pool logic as conditional transactions
+                // to garbage collect transactions out of the bundle range.
+                rollup_args.enable_tx_conditional || builder_args.enable_revert_protection,
+            )
+            .with_supervisor(
```
SozinM: Btw you could remove supervisor, because interop is canceled, missed in https://github.com/flashbots/op-rbuilder/pull/475

akundaz: done

crates/op-rbuilder/src/pool/builder.rs
```diff
@@ -0,0 +1,47 @@
...
+use reth_transaction_pool::{EthPoolTransaction, blobstore::DiskFileBlobStore};
+
+use crate::{args::OpRbuilderArgs, pool::Flashpool, tx::FBPooledTransaction};
+
+pub struct FlashpoolBuilder {
+    op_pool_builder: OpPoolBuilder<FBPooledTransaction>,
+}
+
+impl FlashpoolBuilder {
+    pub fn new(builder_args: &OpRbuilderArgs) -> Self {
```
julio4, line 16: Potentially define FlashpoolArgs and impl From OpRbuilderArgs

julio4 (approved):
This is needed!

PR #496, opened 2026-05-06
chore: use `ActiveHardforks` a bit more widely

akundaz:
Follow up to https://github.com/flashbots/op-rbuilder/pull/493

### DIFF
--- a/crates/op-rbuilder/src/builder/payload_handler.rs
+++ b/crates/op-rbuilder/src/builder/payload_handler.rs
@@ -1,6 +1,7 @@
 use crate::{
     builder::{
-        p2p::Message, payload::build_block, receipt::build_receipt, syncer_ctx::OpPayloadSyncerCtx,
+        OpPayloadJobCtx, p2p::Message, payload::build_block, receipt::build_receipt,
+        syncer_ctx::OpPayloadSyncerCtx,
     },
     evm::OpBlockEvmFactory,
     hardforks::ActiveHardforks,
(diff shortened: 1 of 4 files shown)


crates/op-rbuilder/src/builder/payload_handler.rs
```diff
@@ -305,33 +305,34 @@ where
 
     let evm_factory = OpBlockEvmFactory::new(ctx.evm_config().clone(), evm_env);
 
+    let enable_tx_tracking_debug_logs = ctx.enable_tx_tracking_debug_logs();
+    let max_gas_per_txn = ctx.max_gas_per_txn();
+    let max_uncompressed_block_size = ctx.max_uncompressed_block_size();
+    let job_ctx =
+        ctx.into_op_payload_job_ctx(payload_config, evm_factory, block_env_attributes, cancel);
+
     execute_transactions(
```
avalonche, line 318: do we need to pass in max_gas_per_txn and max_uncompressed_block_size here when it can be accessed via job_ctx?

crates/op-rbuilder/src/builder/payload_handler.rs
```diff
@@ -388,7 +388,7 @@ fn execute_transactions(
         // Note that this *only* needs to be done post-regolith hardfork, as deposit nonces
         // were not introduced in Bedrock. In addition, regular transactions don't have deposit
         // nonces, so we don't need to touch the DB for those.
-        let depositor_nonce = (ctx.is_regolith_active(timestamp) && tx_recovered.is_deposit())
+        let depositor_nonce = (ctx.hardforks.is_regolith_active() && tx_recovered.is_deposit())
```
avalonche, line 384: should https://github.com/flashbots/op-rbuilder/blob/ee65d1abf9f37a78217f58410de880149c31fefe/crates/op-rbuilder/src/builder/payload_handler.rs#L243 be included in the refactor too

crates/op-rbuilder/src/builder/payload_handler.rs
akundaz: fixed

crates/op-rbuilder/src/builder/payload_handler.rs
akundaz: done

julio4 (approved):

Issue #188, opened 2025-07-03
Add Flashblocks RPC module


ferranbt:
The Rbuilder should be able to run the Flashblocks RPC module itself as a extension.

PR #498, opened 2026-05-07
fix: correctly track reverted hashes

akundaz:
## 📝 Summary
Use the new pool wrapper to manage the reverted txs cache. This fixes a bug where tracking the reverted hashes was dependent on the `--builder.log-pool-transactions` flag. This was not caught by our integration tests because the integration tests spin up the tx pool monitor regardless of the value of that flag. 

## 💡 Motivation and Context
The reverted cache should be managed closely to the pool

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/pool/builder.rs
+++ b/crates/op-rbuilder/src/pool/builder.rs
@@ -1,15 +1,24 @@
+use alloy_primitives::TxHash;
+use futures::StreamExt;
+use moka::sync::Cache;
 use reth_evm::ConfigureEvm;
 use reth_node_api::{FullNodeTypes, NodeTypes, PrimitivesTy, TxTy};
 use reth_node_builder::{BuilderContext, components::PoolBuilder};
 use reth_optimism_forks::OpHardforks;
 use reth_optimism_node::OpPoolBuilder;
 use reth_optimism_txpool::{OpPooledTx, OpTransactionPool};
-use reth_transaction_pool::{EthPoolTransaction, blobstore::DiskFileBlobStore};
+use reth_tasks::TaskExecutor;
+use reth_transaction_pool::{
+    AllTransactionsEvents, EthPoolTransaction, FullTransactionEvent, TransactionPool,
+    blobstore::DiskFileBlobStore,
+};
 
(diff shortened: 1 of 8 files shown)


crates/op-rbuilder/src/pool/builder.rs
```diff
@@ -40,8 +52,46 @@ where
...
         evm_config: Evm,
     ) -> eyre::Result<Self::Pool> {
+        let Self {
+            op_pool_builder,
+            enable_revert_protection,
+        } = self;
+
+        let inner_pool = op_pool_builder.build_pool(ctx, evm_config).await?;
+
+        let reverted_cache = enable_revert_protection.then_some(setup_revert_protection(
```
avalonche, line 62: why do we need a reverted cache? is it for easier test assertions?

akundaz: no this is a core feature for revert protection functionality. it's so senders can track if their txs have been evicted from the pool because they were reverted

avalonche: we don't expose rpc methods for transaction receipts though? rollup-boost only forwards bundles and raw transactions

avalonche: if its for debug purposes, the cache size should be configurable - right now the bundle request rate is 300 req/s for reverted transactions on mainnet when the cache max_capacity is only 100

akundaz: this pr is meant to just fix the bug, not add features. i think the cache capacity should be configured in terms of storage size (like configured to a max of 4mb).

and i thought the plan was to expose the rpc for receipts at some point

avalonche: no plans for receipts rpc

avalonche (approved):

crates/op-rbuilder/src/pool/builder.rs
akundaz: this is what i'm referring to https://github.com/flashbots/op-rbuilder/issues/188

PR #499, opened 2026-05-07
refactor: wire presim through flashpool

akundaz:
## 📝 Summary
Refactor the presim startup so that it starts with the pool. Also hook into the `add_transaction` function to rework the presim lifecycle. Previously, txs would be added to the pool and the background presim task would remove them if the tx reverted, now we run the presim task and then add it to the pool if it succeeds. This is now possible because we access the transaction validator from the inner pool and use it to run the several quick checks first.

## 💡 Motivation and Context
Presim should be managed closer to the pool.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/pool/overrides.rs
+++ b/crates/op-rbuilder/src/pool/overrides.rs
@@ -1,0 +1,78 @@
+use std::time::Instant;
+
+use reth_transaction_pool::{
+    AddedTransactionOutcome, PoolResult, PoolTransaction, TransactionOrigin, TransactionPool,
+    TransactionValidationOutcome, TransactionValidator, error::PoolError,
+    pool::AddedTransactionState,
+};
+use tracing::error;
+
+use crate::{
+    pool::{Flashpool, metrics},
+    tx::FBPooledTransaction,
+};
+
+impl<
+    P: TransactionPool<Transaction = FBPooledTransaction> + Clone + 'static,
+    V: TransactionValidator<Transaction = FBPooledTransaction> + Clone,
+> Flashpool<P, V>
+{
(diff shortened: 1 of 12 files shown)


crates/op-rbuilder/src/pool/overrides.rs
```diff
@@ -0,0 +1,78 @@
...
+            .await;
+
+        use TransactionValidationOutcome::*;
+        let transaction = match validation_outcome {
+            Valid { transaction, .. } => transaction,
+            Invalid(tx, err) => return Err(PoolError::new(*tx.hash(), err)),
+            Error(hash, err) => return Err(PoolError::other(hash, err)),
+        };
+
+        if transaction.transaction().is_bundle()
```
avalonche:
```suggestion
        if transaction.transaction().is_revert_protected()
```
we should add to the pool if its a bundle that's not revert protected, otherwise it'll be silently dropped from the pool even when its allowed to be reverted

akundaz: fixed

avalonche (approved):
`tokio::task::spawn_blocking()` in pool/presim.rs uses Tokio's shared blocking pool which is used by the block building pipeline and can starve resources for block building - consider bounding the number of concurrent presim processes or have its own dedicated threads in a different pool

akundaz:
> `tokio::task::spawn_blocking()` in pool/presim.rs uses Tokio's shared blocking pool which is used by the block building pipeline and can starve resources for block building - consider bounding the number of concurrent presim processes or have its own dedicated threads in a different pool

this is a bit involved, will definitely do in a follow up

PR #507, opened 2026-05-12
refactor: use oncelock for payload cancel reason

akundaz:
a bit simpler to manage than a u8 + atomic synchronization


julio4 (approved):
make sense!

PR #514, opened 2026-05-14
feat: chain event task to proxy reth node event logs

julio4:
Spawn a lightweight task that listen to reth_node_events::node engine events and emit debug levels log on target `op_rbuilder::chain` with minimal information (block hash, number, txs)


akundaz:
we already get logs like these, what is this for?

julio4:
> we already get logs like these, what is this for?

to filter most of the fields and disable reth logs and reduce logs volume without losing sequence of events

PR #518, opened 2026-05-15
refactor: extract post-seal side-effects into PostSealHook trait

julio4:
Move ws publication, p2p forwarding, engine propagation, and per-flashblock metrics out of the building loop behind a `PostSealHook` trait. Hooks are constructed once and dispatched after each sealed candidate (fallback + every flashblock).

`build_next_flashblock` no longer performs side-effects: it returns `FlashblockBuildResult` struct and the async loop handles dispatch.

### DIFF
--- a/crates/op-rbuilder/src/builder/hooks/channel.rs
+++ b/crates/op-rbuilder/src/builder/hooks/channel.rs
@@ -1,0 +1,34 @@
+use crate::builder::hooks::post_seal::{PostSealHook, SealedCandidate, SealedCtx};
+use reth_optimism_node::OpBuiltPayload;
+use tokio::sync::mpsc;
+use tracing::warn;
+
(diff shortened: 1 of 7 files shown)


akundaz:
I do want a hook system, but it's not clear to me this is the right design for it. For the time being it would be better to not have the new trait and context types, and instead extract the side-effects a different way where each one only has the minimal amount of information for it to run.

julio4:
> I do want a hook system, but it's not clear to me this is the right design for it. For the time being it would be better to not have the new trait and context types, and instead extract the side-effects a different way where each one only has the minimal amount of information for it to run.

Hm yes I agree that the ctx is encapsulating too much here. I tried to rework this by adding a more immutable Slot Metadata and derive more information directly from the payload built itself.

avalonche (approved):

PR #522, opened 2026-05-20
feat(presim): limit max concurrency

akundaz:
## 📝 Summary
Simple presim concurrency limiting using a semaphore. Configurable via the `--builder.presim-max-concurrent` flag.

## 💡 Motivation and Context
We want to limit the max number of presim tasks running so we don't starve the tokio executor.

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -72,6 +72,14 @@ pub struct OpRbuilderArgs {
     )]
     pub pre_simulate_bundles: bool,
 
+    /// The maximum number of concurrent presim tasks to run.
+    #[arg(
+        long = "builder.presim-max-concurrent",
+        default_value = "4",
+        env = "PRESIM_MAX_CONCURRENT"
+    )]
+    pub presim_max_concurrent: usize,
+
     /// Enables logs to trace transaction lifecycle as it is added to the
     /// mempool and added to a block. Requires `RUST_LOG=tx_trace=debug` in
     /// addition to this flag.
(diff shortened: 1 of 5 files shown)


crates/op-rbuilder/src/args/op.rs
```diff
@@ -72,6 +72,14 @@ pub struct OpRbuilderArgs {
     )]
     pub pre_simulate_bundles: bool,
 
+    /// The maximum number of concurrent presim tasks to run.
+    #[arg(
+        long = "builder.presim-max-concurrent",
+        default_value = "4",
+        env = "PRESIM_MAX_CONCURRENT"
+    )]
+    pub presim_max_concurrent: usize,
```
julio4: Add a non-zero assertion presim_max_concurrent > 0 else it can causes deadlock, or set 0 to bypass limit (unlimited)

julio4 (dismissed):
LGTM, just details

crates/op-rbuilder/src/args/op.rs
akundaz: done

crates/op-rbuilder/src/args/op.rs
```diff
@@ -72,6 +72,15 @@ pub struct OpRbuilderArgs {
     )]
     pub pre_simulate_bundles: bool,
 
+    /// The maximum number of concurrent presim tasks to run. If omitted,
+    /// concurrency will be unbounded.
+    #[arg(
+        long = "builder.presim-max-concurrent",
+        default_value = "4",
```
avalonche: why 4 as default? we get 200 reqs per second on mainnet

akundaz: i don't want to leak configuration info

avalonche: then there shouldn't be a default value - having this means when the flag is omitted presim_max_concurrent is always Some()

akundaz: removed

avalonche:
no tests for new sephamore path?

avalonche (approved):

akundaz:
> no tests for new sephamore path?

added unit test

PR #523, opened 2026-05-20
feat: add reverted_tx_count counter metric

julio4:
Replaces the per-tx `skipping reverted transaction` log (already at `trace!`) with an explicit `op_rbuilder_reverted_tx_count` Counter so reverted-tx volume is observable without flooding logs.


akundaz:
this conflicts with https://github.com/flashbots/op-rbuilder/pull/492, can you work together on this?

PR #525, opened 2026-05-21
fix: don't publish flashblock after payload cancellation

avalonche:
## 📝 Summary

Prevents a cancelled payload job from publishing locally built payloads into the payload handler after the job has already been resolved (by newPayload). 

It also renames `PayloadJobCancellation::cancelled()` to `wait_for_cancellation()` so async wait sites are easier to distinguish from immediate `is_cancelled()` checks.

## 💡 Motivation and Context

1. `forkchoiceUpdated` starts a payload job.
2. The job builds fallback and flashblock payloads.
3. Built payloads are sent through local channels to the `PayloadHandler`.
4. The handler emits `Events::BuiltPayload`, which updates reth's engine tree with the local payload.

`build_next_flashblock()` was a blocking job that previously performed side effects such as publishing. Reth could then see a stale local same-height payload after the job had already been resolved. Now:

- `build_next_flashblock()` builds and returns `BuiltFlashblockOutput`.
- The async outer loop owns publication.
- The outer loop checks cancellation after the blocking task returns and before publishing.
- `publish_flashblock_payload()` performs one final cancellation check before websocket publish and channel sends.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/builder/payload.rs
+++ b/crates/op-rbuilder/src/builder/payload.rs
@@ -4,7 +4,7 @@ use crate::{
         BuilderConfig,
         best_txs::{FlashblockPoolTxCursor, FlashblockTxTracker},
         builder_tx::{BuilderTransactions, reserve_builder_tx_budget},
-        cancellation::FlashblockJobCancellation,
+        cancellation::{FlashblockJobCancellation, PayloadJobCancellation},
         context::{OpPayloadBuilderCtx, OpPayloadJobCtx},
         generator::{BuildArguments, PayloadBuilder},
         timing::{FlashblockScheduler, compute_slot_offset_ms},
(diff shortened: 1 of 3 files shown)


julio4 (commented):
We were already checking for block cancel before calling all publish side effects in sync, so this is mostly refactor to extract publish side effects outside of build_next_flashblock it seems?

julio4:
Also please see #518

crates/op-rbuilder/src/builder/payload.rs
```diff
@@ -900,6 +927,87 @@ where
...
+        built_flashblock: BuiltFlashblockOutput,
+    ) -> eyre::Result<Option<FlashblocksState>> {
+        let BuiltFlashblockOutput {
+            next_flashblock_state,
+            new_payload,
+            fb_payload,
+            build_duration,
+        } = built_flashblock;
+
+        if payload_cancel.is_cancelled() {
```
akundaz, line 964: Aren't we checking for cancellation just before we call this publish function? So this seems unnecessary here

avalonche: I believe we check is_resolved but is_cancelled() can also be true if deadline / error is reached. but yes I believe it can be refactored better

akundaz (approved):

PR #526, opened 2026-05-21
refactor: start cleaning up execution info

akundaz:
Just a few small refactors in this PR:
- cleaning up limit checking
- moving last fb index from fb state type into execution info

Budgeting refactor still in progress


crates/op-rbuilder/src/execution.rs
```diff
@@ -231,14 +299,29 @@ fn build_receipt<E: Evm>(
 
 #[cfg(test)]
 mod tests {
+    use alloy_primitives::U256;
+
     use super::{ExecutionInfo, TxnExecutionResult};
 
+    fn execution_info_with_uncompressed_bytes(cumulative_uncompressed_bytes: u64) -> ExecutionInfo {
```
julio4, line 306: nit: you can also add a implementation on ExecutionInfo under test cfg to do ExecutionInfo::with_uncompressed_bytes

julio4 (commented):
I don't think ExecutionInfo is the right place to keep track of flashblocks index. I would prefer if possible splitting in object of same lifecycle: one for the slot/payload, one per flashblock.

budget check functions are nice but we could do even more, what do you have in mind for "Budgeting refactor still in progress"?

Rest looks good!

PR #528, opened 2026-05-25
refactor(payload): introduce FlashblockMeta

julio4:
Group together single piece of data (derived from flashblock state) in the type system with FlashblockMeta.

`BuilderTransactions` trait still takes the two bools but a follow-up can swap the trait arg to `&FlashblockMeta` to get a self-consistent cursor flowing through the builder-tx API instead of a derived-bool pair.

### DIFF
--- a/crates/op-rbuilder/src/builder/payload.rs
+++ b/crates/op-rbuilder/src/builder/payload.rs
@@ -196,12 +196,11 @@ impl FlashblocksState {
         self.target_flashblock_count
     }
 
-    fn is_first_flashblock(&self) -> bool {
-        self.flashblock_index == 0
-    }
-
-    fn is_last_flashblock(&self) -> bool {
-        self.flashblock_index == self.target_flashblock_count
+    fn meta(&self) -> FlashblockMeta {
+        FlashblockMeta {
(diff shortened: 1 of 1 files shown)


crates/op-rbuilder/src/builder/payload.rs
```diff
@@ -234,6 +233,24 @@ impl FlashblocksState {
...
 
+/// Projection of [`FlashblocksState`] describing where the current
+/// build sits within the slot's flashblock sequence.
+#[derive(Debug, Clone, Copy)]
+struct FlashblockMeta {
+    flashblock_index: u64,
+    target_flashblock_count: u64,
+}
+
+impl FlashblockMeta {
```
avalonche, line 244: why not include other fields like the target count?

julio4: The idea is that FlashblockMeta is a projection of FlashblocksState only for BuilderTx. And BuilderTx acts as some kind of pre-seal 'hook', so this projection can be expanded later on if needed but kept as minimal as possible now.

avalonche (approved):

PR #530, opened 2026-06-01
chore: bump op-reth v2.2.5

avalonche:
## 📝 Summary

Upgrade to op-reth v2.2.5


crates/op-rbuilder/src/pool/overrides.rs
```diff
@@ -12,33 +11,17 @@ use crate::{
...
-    V: TransactionValidator<Transaction = FBPooledTransaction> + Clone,
-> Flashpool<P, V>
-{
+impl<P: TransactionPool<Transaction = FBPooledTransaction> + Clone + 'static> Flashpool<P> {
     pub(super) async fn add_transaction_override(
         &self,
         origin: TransactionOrigin,
         transaction: FBPooledTransaction,
     ) -> PoolResult<AddedTransactionOutcome> {
-        let validation_outcome = self
```
akundaz, line 25: Why did you remove the validator? It's necessary for presim because we need to do the fast checks before spending the cpu cycles on simulation

avalonche: hmm reth v2 removed the `validator()` access from `inner_pool()`, let me try add it back upstream

julio4 (commented):
Not needed after #532

PR #537, opened 2026-06-12
Make op-rbuilder compatible with upcoming op-node

joshklop:
## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args/op.rs
+++ b/crates/op-rbuilder/src/args/op.rs
@@ -41,7 +41,15 @@ pub struct OpRbuilderArgs {
     pub max_gas_per_txn: Option<u64>,
 
     /// Maximum cumulative uncompressed (EIP-2718 encoded) block size in bytes
-    #[arg(long = "builder.max-uncompressed-block-size")]
+    // Explicit `id` avoids a clap argument-name collision: as of op-reth
+    // v2.3.2-rc.2, the flattened `RollupArgs` also defines a field named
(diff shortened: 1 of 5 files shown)


julio4 (approved):

crates/op-rbuilder/src/args/op.rs
```diff
@@ -41,7 +41,15 @@ pub struct OpRbuilderArgs {
     pub max_gas_per_txn: Option<u64>,
 
     /// Maximum cumulative uncompressed (EIP-2718 encoded) block size in bytes
-    #[arg(long = "builder.max-uncompressed-block-size")]
+    // Explicit `id` avoids a clap argument-name collision: as of op-reth
+    // v2.3.2-rc.2, the flattened `RollupArgs` also defines a field named
+    // `max_uncompressed_block_size` (`--rollup.max-uncompressed-block-size`),
```
avalonche, line 46: Is this the same functionality as what's in op-reth? This flag should be mark deprecated

avalonche (approved):

avalonche:
> Commits must have verified signatures.

crates/op-rbuilder/src/args/op.rs
joshklop: Should be same functionality but the op-reth flag might not be used in op-rbuilder, depending on what upstream functionality op-rbuilder uses. Does that make sense?

PR #542, opened 2026-06-16
test(continuous): per-instance trigger-miss hook

julio4:
Fallback test was flaky with the force trigger-miss hook, that relied on a process-global count that any concurrent test could consume. Replaced with per-builder Arc<AtomicU64>.


crates/op-rbuilder/src/args/op.rs
```diff
@@ -251,6 +251,11 @@ pub struct FlashblocksArgs {
         default_value = "false"
     )]
     pub flashblocks_continuous_build: bool,
+
+    /// Number of forced `SharedBest::take()` misses for test isolation.
+    /// Consumed only under `#[cfg(test)]`; else always 0.
+    #[arg(skip)]
+    pub initial_force_take_miss_count: u64,
```
avalonche: can this not be refactored such that the test instance can be configured with a test SharedBest instance with test config?

PR #549, opened 2026-07-07
refactor: builder transactions overhaul

akundaz:
# Summary
Major overhaul of builder transaction management.

## Motivation and Context

Previously, builder transaction code:
- Expressed position expressed as flags: every call site threaded `top_of_block`, `is_first_flashblock`, `is_last_flashblock` bools through the `BuilderTransactions` trait, each impl re-derived its behavior from flag combinations, and an `is_top_of_block: bool` on the produced tx decided at commit time what actually landed.
- Flashtestations was nested inside the flashblocks builder-tx types. This makes adding or repositioning a builder tx awkward and error-prone.

This refactor gets us closer to exposes builder transaction functionality as a library, i.e. if other builders want to have additional builder transactions this should make it easier.

# What changed

## Behavior changes

Previously if the flashblocks number contract tx failed, we would add another claim tx. Now it falls back to not writing anything at all, since the claim tx should be independent of the flashblocks number contract. This affected a lot of the integration tests.

## Positional dispatch

Position is now a type, replacing flag threading, and is declared once at startup time:

- `BuilderTxPosition` enum (`TopOfBlock` / `TopOfFlashblock` / `BottomOfBlock`) on a `ScheduledBuilderTx` entry pairs each producer with its dispatch point.
- `BuilderTxSchedule` owns the entries and exposes one method per dispatch point (`commit_top_of_block`, `commit_top_of_flashblock`, `estimate_bottom_of_block`, `commit_bottom_of_block`). The payload builder is responsible for calling the right method at the right moment; no position flags cross the trait boundary.
- Flashtestations is untangled: it's an independent `BottomOfBlock` schedule entry rather than a field inside the flashblocks types. Nonce chaining between entries is done uniformly by the schedule (each entry's txs are committed to the shared simulation state before the next entry runs).

## Module split

`builder_tx.rs` → `builder_tx/` with one concept per file: `env.rs` (`BuilderTxEnv`), `sim.rs` (simulation state + helpers), `producer.rs` (trait, artifact, error), schedule.rs (dispatch), `impls/claim.rs`, `impls/flashblock_number.rs`.

## Naming overhaul

"Builder tx" previously meant three different things (the on-chain tx, the component producing it, and the collection of producers). Now: `BuilderTxProducer` produces `SimulatedBuilderTx` values, dispatched by a `BuilderTxSchedule`.

The block claim tx (the tx from the builder's address with the "Block Number: {}" message) is named for what it is: `ClaimBuilderTx`, replacing the misleading "fallback"/"base" naming.

Error/type spellings converge on Tx (`BuilderTxError`), and `ctx: &BuilderTxEnv` parameters are now `env` throughout.

## Fixes along the way:

- The generic `BuilderTx` type parameter is gone from `OpPayloadBuilder`.
- The producer trait shrank from 7 parameters to 4, and ExecutionInfo is now a `&` (was `&mut`).
- Bottom-of-block dispatch now only runs on the last flashblock, removing a wasted simulation per flashblock.

# Future work

The public API is fairly large (see `crates/op-rbuilder/src/builder/mod.rs`). Some work will need to go into redesigning this so it's suitable as an rblib feature.


crates/op-rbuilder/src/builder/builder_tx/producer.rs
```diff
@@ -0,0 +1,92 @@
...
+    TransactionHalted(Address, OpHaltReason),
+    #[error("transaction to {0} reverted {1}")]
+    TransactionReverted(Address, Revert),
+    #[error("invalid transaction error {0}")]
+    InvalidTransactionError(Box<dyn core::error::Error + Send + Sync>),
+    #[error("evm execution error {0}")]
+    EvmExecutionError(Box<dyn core::error::Error + Send + Sync>),
+    #[error(transparent)]
+    Other(Box<dyn core::error::Error + Send + Sync>),
+}
```
julio4, line 48: Is there not alloy types for these?

crates/op-rbuilder/src/builder/builder_tx/impls/claim.rs
```diff
@@ -0,0 +1,109 @@
...
+        });
+
+        let zero_cost = zero_bytes * 4;
+        let nonzero_cost = nonzero_bytes * 16;
+
+        let tokens_in_calldata = zero_bytes + nonzero_bytes * 4;
+        let floor_gas = 21_000 + tokens_in_calldata * TOTAL_COST_FLOOR_PER_TOKEN;
+
+        std::cmp::max(zero_cost + nonzero_cost + 21_000, floor_gas)
+    }
```
julio4, line 82: Seems more like min gas limit than estimate tx gas.

Also consider:
```rust
let intrinsic_gas = 21_000
    + input
        .iter()
        .map(|byte| if *byte == 0 { 4 } else { 16 })
        .sum::<u64>();

let floor_gas = eip7623::transaction_floor_cost(eip7623::tokens_in_calldata(input));

intrinsic_gas.max(floor_gas)
```

crates/op-rbuilder/src/builder/builder_tx/impls/flashblock_number.rs
```diff
@@ -0,0 +1,216 @@
...
+            self.signed_increment_flashblocks_tx(env, &mut evm)
+        };
+
+        let tx = num_tx.inspect_err(|e| {
+            warn!(
+                target: "builder_tx",
+                error = %e,
+                "flashblocks number contract tx simulation failed"
+            )
+        })?;
```
julio4: Consider a `flashblock_number_tx_failure_total` metric

julio4 (approved):
Good isolation of builder tx!

avalonche:
> Previously if the flashblocks number contract tx failed, we would add another claim tx. Now it falls back to not writing anything at all, since the claim tx should be independent of the flashblocks number contract. 

The flashblocks number contract tx had two functionalities:
- we are able to tell whether the builder built this flashblock
- incrementing the flashblocks number contract

If the fallback tx is removed, how do we know this flashblock was built by the builder? In this change, plain flashblocks without the flashblock number contract + flashtestations (most l2 chains) will be missing the builder tx in the last flashblock. 

The first top of block builder tx was added because of the fcu avalanche issue and can be removed if its resolved on op-node.

akundaz:
> > Previously if the flashblocks number contract tx failed, we would add another claim tx. Now it falls back to not writing anything at all, since the claim tx should be independent of the flashblocks number contract.
> 
> The flashblocks number contract tx had two functionalities:
> 
> * we are able to tell whether the builder built this flashblock
> * incrementing the flashblocks number contract
> 
> If the fallback tx is removed, how do we know this flashblock was built by the builder? In this change, plain flashblocks without the flashblock number contract + flashtestations (most l2 chains) will be missing the builder tx in the last flashblock.
> 
> The first top of block builder tx was added because of the fcu avalanche issue and can be removed if its resolved on op-node.

I've reverted the behavior changes, though to maintain the original behavior, there's a claim tx added to the bottom of the block if the flashblock number contract isn't enabled, which I'm not sure yet if we intended in the past

PR #556, opened 2026-07-10
safety: supervise flashblock scheduler task

julio4:
Spawns the flashblock scheduler through task executor.

Also catches panics (flashblock_scheduler_death counter). Previously dropped trigger channel silently and could look like a completed schedule. This is a safety measure as scheduler should never panic.

### DIFF
--- a/crates/op-rbuilder/src/builder/payload.rs
+++ b/crates/op-rbuilder/src/builder/payload.rs
@@ -20,6 +20,7 @@ use crate::{
 };
 use alloy_primitives::B256;
 use eyre::WrapErr as _;
+use futures::FutureExt;
 use op_alloy_rpc_types_engine::OpFlashblockPayload;
 use reth_chainspec::EthChainSpec;
 use reth_node_api::PayloadBuilderError;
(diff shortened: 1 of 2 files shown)


akundaz:
Why is this safety measure needed? I don't know of any source of panicking with that task

julio4:
> Why is this safety measure needed? I don't know of any source of panicking with that task

It's not needed but nice to have; rust already enforce handling all cases including panics, but here the task is detached so the failure signal is the same as the success one (recv None). There's no per job overhead, and it's good practice to atleast catch panic on detached task.

akundaz:
> > Why is this safety measure needed? I don't know of any source of panicking with that task
> 
> It's not needed but nice to have; rust already enforce handling all cases including panics, but here the task is detached so the failure signal is the same as the success one (recv None). There's no per job overhead, and it's good practice to atleast catch panic on detached task.

It's adding complexity. I understand what you're saying that in general it's good to be safe but this is a code path we should just guarantee never panics.

julio4:
> > > Why is this safety measure needed? I don't know of any source of panicking with that task
> > 
> > 
> > It's not needed but nice to have; rust already enforce handling all cases including panics, but here the task is detached so the failure signal is the same as the success one (recv None). There's no per job overhead, and it's good practice to atleast catch panic on detached task.
> 
> It's adding complexity. I understand what you're saying that in general it's good to be safe but this is a code path we should just guarantee never panics.

I think the complexity is scoped but I understand, one improvement could be to separate metrics in a new OpRBuilderDebugMetrics. We can remove it I don't hold strong opinion, I just tried to pinpoint failure points as I would like to try some refactoring around flashblocks scheduling. Feel free to close.

PR #557, opened 2026-07-10
safety(continuous): pool-epoch listener liveness metrics

julio4:
Adds pool_change_epoch gauge + pool_epoch_listener_exit counter so we can detect dead pool-change listener from a genuinely idle pool. A dead listener could silently degrades continuous candidate gating.

### DIFF
--- a/crates/op-rbuilder/src/builder/service.rs
+++ b/crates/op-rbuilder/src/builder/service.rs
@@ -16,6 +16,7 @@ use crate::{
     traits::{NodeBounds, PoolBounds},
 };
 use eyre::WrapErr as _;
+use futures::FutureExt;
 use reth_node_api::NodeTypes;
 use reth_node_builder::{BuilderContext, components::PayloadServiceBuilder};
 use reth_optimism_evm::OpEvmConfig;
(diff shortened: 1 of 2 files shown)


crates/op-rbuilder/src/builder/service.rs
```diff
@@ -121,11 +122,22 @@ impl FlashblocksServiceBuilder {
...
-                    pool_change_epoch.fetch_add(1, Ordering::Relaxed);
-                }
+                let listen = async {
+                    // Every new pending tx may improve current candidate
+                    while pending_txs.recv().await.is_some() {
+                        let epoch = pool_change_epoch.fetch_add(1, Ordering::Relaxed) + 1;
+                        listener_metrics.pool_change_epoch.set(epoch as f64);
+                    }
+                };
+                let _ = std::panic::AssertUnwindSafe(listen).catch_unwind().await;
```
akundaz, line 134: Why do you need to catch panics for this?

julio4: The counter is an alert signal, without catching panic we would skip it and miss alert.

akundaz: I mean have you been seeing anything to motivate tracking this? This closure is very simple so I don't see the need to catch a panic here

julio4: I would like to move all this pool logic somewhere else, so from the builder service I consider this closure like a black box even if it is very simple (at the moment).

Now if the pool listener panic for whatever reason we can't tell anything from the builder service, and can't make the difference between an empty pool or a panicked one. But totally agree that in the current shape this seems useless, I can always add this later with more changes that justify this if needed.

PR #561, opened 2026-07-20
test: karst coverage

julio4:
Adds `karst_node_config()` to in-process test framework 

Some tests covering:
- karst-at-genesis block building 
- P256VERIFY precompile 
- P256VERIFY Osaka gas repricing


avalonche (approved):
It'll be good to update builder-playground too

PR #562, opened 2026-07-23
feat: support decker devnet artifacts

julio4:
Extends `--builder.playground` to recognize decker runtime/artifacts/ layout.

For now keeping builder-playground support as legacy option but can be removed later.


akundaz (approved):
lgtm, but what is decker?

the ai also seems to have caught a couple of typos

julio4:
> lgtm, but what is decker?
> 
> the ai also seems to have caught a couple of typos

Its the new builder playground, more lean and more friendly for agents.
https://github.com/flashbots/decker

PR #579, opened 2026-09-23
fix: remove max_gas_per_txn check in syncer path

akundaz:
Deposits should be force-included from L1 and not subject to per-transaction gas limits. This was only being followed in the building path, not the syncer.


crates/op-rbuilder/src/builder/payload_handler.rs
sebastianst:
```suggestion
```
I'd just remove this entire check, the flag is a builder flag only. The verifier path shouldn't enforce a gas per tx limit if the tx is otherwise valid.

akundaz: Fair enough, removed the check

avalonche (approved):

